PR #126, opened 2023-09-19
ci: use a fork for denobot bumping versions

dsherret:
This way we don't need to give denobot write permissions to this repo.


.github/workflows/update_versions.ts
```diff
@@ -12,16 +12,16 @@ interface ReplacementsData {
 }
 
 const replacementsFile = $.path("replacements.json");
```
dsherret, line 14: Oh, I think I can try running this before merging. One second.

dsherret:
Going to bypass branch protections in the interest of time in order to iterate and test this out. Seems I can't run this properly when this is on a branch.

PR #130, opened 2023-09-20
fix: replace og-image

### LINKED ISSUE #16: Create new social / OG image for Deno Docs
Need a new dedicated social image that is docs-specific (ideally). Should replace current `static/social.png`.

hashrock:
Closes #16


kwhinnery (approved):
thanks @hashrock

PR #131, opened 2023-09-20
docs: Correct code example for `kv` variable in manual

mohsinalipro:
This commit addresses a minor issue in the Deno documentation manual related to the `kv` variable. The code example provided had an incorrect reference to the variable used to open the key-value store. The variable should be `kv`, not `db`.

**Changes Made:**

- Corrected the variable name from `db` to `kv` in the code example.
- Verified the correctness of the code snippet.


kwhinnery (approved):
thank you!

PR #132, opened 2023-09-21
fix: add sidebar for runtime root page

### LINKED ISSUE #94: Add sidebar for runtime root page
https://docs.deno.com/runtime/manual/runtime

bipo38:
Closes #94


kwhinnery:
Thanks for looking into this one, but I think it might be better to just redirect that URL - the page doesn't have a lot of value by itself. I just added in a redirect for it, so we should be good. Thanks again!

PR #135, opened 2023-09-21
Include more VS Code instructions in jupyter.md

rgbkrk:
Documentation from providing support in https://github.com/denoland/deno/issues/20600

cc @guy-borderless


kwhinnery:
thank you!

PR #137, opened 2023-09-22
fix: Replace "Back to main menu" navigation on mobile Navbar to provide clearer context

### LINKED ISSUE #52: Explore better alternatives for "Back to main menu" product-level navigation on mobile
Right now, the mobile nav doesn't provide a lot of context for where you are in the docs.deno.com hierarchy for our different "products" (Runtime, Deploy, and KV). 

The ideal experience would be that the menu shows "You are in the docs for Runtime", and then makes it obvious that you can switch to another set of docs in some way. Example:

image

hashrock:
Closes #52 

before:

image

after:

https://github.com/denoland/deno-docs/assets/3132889/31994e0b-443b-4670-9760-8af3a048ec16

This UI change does not affect the desktop layout.


kwhinnery (approved):

kwhinnery:
I reintroduced the home page and tweaked a few bits, but it's looking good - thanks @hashrock!

PR #138, opened 2023-09-23
fix: variable name in the example of kv get function

IndranilKhedkar:
Update index.mdx - Correcting the variable name in the example of kv get function


hashrock (approved):
Thank you!

PR #141, opened 2023-09-24
update: workers now work in compiled executables

Leokuma:
https://discord.com/channels/684898665143206084/1151939558548709486/1151943237339197450
> Andreu Botella (they/them)
 — 
14/09/2023 15:11
it used to be that you couldn't use workers, but now you can if you pass the `--include` flag with the worker scripts you will use
I guess the documentation is stale on that


hashrock (approved):
checked with local env. LGTM!

PR #142, opened 2023-09-25
fix: Changed -NaN to NaN 

IndranilKhedkar:
Close https://github.com/denoland/deno-docs/issues/76 
Update key_space.mdx

### DIFF
--- a/kv/manual/key_space.mdx
+++ b/kv/manual/key_space.mdx
@@ -52,7 +52,7 @@ Within a given type, the ordering is:
 
 - `Uint8Array`: byte ordering of the array
 - `string`: byte ordering of the UTF-8 encoding of the string
-- `number`: -NaN < -Infinity < -1.0 < -0.5 < -0.0 < 0.0 < 0.5 < 1.0 < Infinity <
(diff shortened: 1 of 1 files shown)


hashrock:
I think NaN shouldn't appear twice because it's about key ordering

kt3k:
cc @losfair Is NaN bigger than `Infinity`, or less than `-Infinity`?

IndranilKhedkar:
As per Lodash `sortBy` function `NaN` is bigger than `Infinity`
image

kt3k:
I also checked with the below script:
```js
const kv = await Deno.openKv(":memory:");
await kv.set([Infinity], 1);
await kv.set([NaN], 1);
await kv.set([-Infinity], 1);
for await (const res of kv.list({ prefix: [] })) {
  console.log(res);
}
```
It prints the below:
```
{ key: [ -Infinity ], value: 1, versionstamp: "00000000000000030000" }
{ key: [ Infinity ], value: 1, versionstamp: "00000000000000010000" }
{ key: [ NaN ], value: 1, versionstamp: "00000000000000020000" }
```
So it looks like `Infinity < NaN` in Deno KV

kt3k (approved):
LGTM

PR #152, opened 2023-09-29
Customize docs breadcrumbs

josh-collinsworth:
- Shows breadcrumbs at all times (unless the user is on a landing page and hasn't begun to navigate any deeper yet)
- Changes the "home" link in the docs to go to the current section's home (`/kv`, `/runtime`, or `/deploy`, as the case may be)
- Slightly modifies styling of breadcrumbs

image

CleanShot 2023-09-29 at 10 58 03@2x

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -99,11 +99,6 @@ h3 {
   padding: 60px 0;
 }
 
-/* Hide breadcrumbs */
-.theme-doc-breadcrumbs {
-  display: none;
-}
-
 .theme-doc-sidebar-container {
   position: relative;
 }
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):
Nice!

josh-collinsworth:
Hmmm now that I'm looking at this, though…I think there's a missing item there. 🤔

kwhinnery:
In terms of visuals and behavior I think this is what we want, but we will need to make these under pages actually do something now because the client-side navigation isn’t redirecting /kv to /kv/manual (for example) as is happening in prod right now. We could move the quick start for each product to this path and redirect the manual index page.

josh-collinsworth:
@kwhinnery I noticed that the KV "home" route said it redirected in prod; I assumed that was the case for all of them. I guess I should have checked.

Would that concept where everything is moved to an accordion help here, or does that still have the same issue? Mainly just curious.

kwhinnery:
WRT the redirects - Docusaurus does client-side routing for the pages it can detect in the folders containing the content. Previously, we didn't have any Docusaurus-generated client-side links to the home pages for any of /kv, /runtime, or /deploy, so the only time those routes would have been requested was when the server had to try and render those pages. And then on the server, we redirected /kv to /kv/manual, etc.

But now that we have a solution for breadcrumbs and have started manually specifying the sidebar contents, we can move the "quick start" content back to the base routes for /kv, /runtime, and /deploy. Will update your branch with this.

kwhinnery:
Going to fix the hover state real quick also: 
image

PR #175, opened 2023-11-01
Updates for version 1.38

bartlomieju:


runtime/manual/tools/documentation_generator.md
```diff
@@ -25,6 +25,52 @@ function add(x: number, y: number): number
...
 ```
 
+## Linting
+
+You can use `--lint` flag to check for problems in your documentation.
+`deno doc` will point out missing JSDoc comments and return types for public
+APIs.
+
+```js
+// multiply.ts
```
dsherret, line 35: Might be best to rename this `mod.ts` so people get the idea to provide the entrypoint.

kwhinnery: also you can provide a title in code blocks if you want rather than a comment: https://docusaurus.io/docs/markdown-features/code-blocks#code-title

dsherret: Oh, cool!

PR #179, opened 2023-11-02
chore: update skypack imports to npm: imports

GJZwiers:
I wrote a code example a long time ago that uses Skypack URLs. Because that no longer seems to be actively maintained I updated them to npm: imports, which also works.

### DIFF
--- a/runtime/manual/basics/testing/index.md
+++ b/runtime/manual/basics/testing/index.md
@@ -453,7 +453,7 @@ that provides test spies and can be used in Deno by importing it from a CDN,
 such as Skypack:
 
 ```js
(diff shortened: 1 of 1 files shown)


runtime/manual/basics/testing/index.md
```diff
@@ -453,7 +453,7 @@ that provides test spies and can be used in Deno by importing it from a CDN,
 such as Skypack:
```
kwhinnery, line 453: Looks like the line above still references Skypack?

GJZwiers: @kwhinnery Thanks, I have updated the line to mention NPM instead.

PR #221, opened 2023-12-07
add kv watch

lambtron:
i kind of copied code from our API reference (https://deno.land/api@v1.38.5?unstable=&s=Deno.Kv&p=prototype.watch) so feel free to zhuzh it up so it matches the level of technical fidelity of the surrounding documentation.


lambtron:
yeah seems like it. not sure what happened there

PR #238, opened 2023-12-20
Add link to install necessary tools for contributing

chengr4:
As a beginner, I didn't even know I needed `--recurse-submodules` to clone project and to install many tools before running `cargo test`. This prompt can prevent beginners like me from feeling discouraged right from the start.

This may related to issue: https://github.com/denoland/deno/issues/21652

### DIFF
--- a/runtime/manual/references/contributing/index.md
+++ b/runtime/manual/references/contributing/index.md
@@ -32,6 +32,9 @@ While iterating on such modules it is recommended to include
 development mode where the JS/TS sources are not included in the binary but read
 at runtime, meaning the binary will not have to be rebuilt if they are changed.
 
+> To use the commands below, you need to first install the necessary tools on
+> your system as described [here](building_from_source).
+
 ```sh
 # cargo build
 cargo build --features __runtime_js_sources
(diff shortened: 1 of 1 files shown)


kwhinnery:
Thanks for this contribution! I assume this was added because you can't follow the cargo commands listed below without installing the tooling first - if so, I would add something like this just above the cargo commands instead. You could say:

"To use the commands below, you need to first install the necessary tools on your system as described here<link>."

Also, rather than hard coding a production URL, please use a relative path, which will allow for testing the link in local development more easily.

chengr4:
Hi @kwhinnery,

Thanks for replying.

At [Contributing page](https://docs.deno.com/runtime/manual/references/contributing/), there are a few places contain the cargo commands. I assume you would like me to modify them as follows:

In `deno` section above code block

> To use the commands below, you need to first install the necessary tools on your system as described [here](relative path)

```sh
# cargo build
cargo build --features __runtime_js_sources

# cargo run -- run hello.ts
cargo run --features __runtime_js_sources -- run hello.ts

# cargo test integration::node_unit_tests::os_test
cargo test --features __runtime_js_sources integration::node_unit_tests::os_test
```

In Submitting a PR to `deno` section before the first number

> To use the commands below, you need to first install the necessary tools on your system as described [here](relative path)

1. `cargo test` passes - this will run full test suite for deno including unit tests, integration tests and Web Platform Tests

---

I hope I understand correctly.

kwhinnery:
That's correct, I think that would help make it obvious why that step is necessary.

chengr4:
@kwhinnery I fixed the commit. When you have time, please take a look. Thank you.

kwhinnery (approved):

kwhinnery:
thanks, lgtm!

PR #262, opened 2024-01-09
Update building_from_source.md

muddlebee:
[markdown rendered](https://github.com/muddlebee/deno-docs/blob/patch-1/runtime/manual/references/contributing/building_from_source.md)

**context**: faced some challenges manually building deno from scratch in WSL, https://github.com/denoland/deno/issues/21865#event-11430554355

### DIFF
--- a/runtime/manual/references/contributing/building_from_source.md
+++ b/runtime/manual/references/contributing/building_from_source.md
@@ -10,7 +10,7 @@ chapter).
 > Deno uses submodules, so you must remember to clone using
 > `--recurse-submodules`.
 
-**Linux**/**Mac**:
+**Linux**/**Mac**/**WSL**:
 
 ```shell
 git clone --recurse-submodules https://github.com/denoland/deno.git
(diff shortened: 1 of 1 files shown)


runtime/manual/references/contributing/building_from_source.md
```diff
@@ -76,6 +76,15 @@ brew install llvm
 # Add /opt/homebrew/opt/llvm/bin/ to $PATH
 ```
 
+**WSL**
+```shell
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+./llvm.sh 16
+```
```
marvinhagemeister, line 84: More out of curiosity: Why does llvm need to be installed this way rather than via `apt/apt-get`? Is the default llvm version in ubuntu too old?

muddlebee: Yes. I faced issues installing clang 16 which is not readily available in Ubuntu 22.04 LTS repos.

folks have raised issues at stack overflow here https://askubuntu.com/questions/1473403/how-to-install-clang-16-on-22-04-lts

runtime/manual/references/contributing/building_from_source.md
```diff
@@ -76,6 +76,15 @@ brew install llvm
...
+
+apt install --install-recommends -y cmake libglib2.0-dev
```
marvinhagemeister, line 86: Let's include this line in the code block

````suggestion
apt install --install-recommends -y cmake libglib2.0-dev
```
````

marvinhagemeister (approved):
Nice, thanks for updating the docs for WSL 👍

PR #265, opened 2024-01-11
Visual/color updates, and a homepage refresh

josh-collinsworth:


src/css/custom.css
```diff
@@ -7,42 +7,79 @@
...
-  --ifm-color-primary-darkest: #205d3b;
-  --ifm-color-primary-light: #33925d;
-  --ifm-color-primary-lighter: #359962;
-  --ifm-color-primary-lightest: #3cad6e;
-  --ifm-code-font-size: 95%;
-  --ifm-code-padding-horizontal: 0.5em;
-  --ifm-code-padding-vertical: 0.25em;
-  --ifm-navbar-search-input-background-color: #fff;
-  --docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.1);
-  --ifm-font-family-base: Inter, ui-sans-serif, system-ui, -apple-system,
```
donjo, line 28: Did you mean to have the font family cut out? It's rendering the system font stack right now but I'm down to see what it looks like with what we chose on the marketing pages

kwhinnery:
These columns probably want to be equal height? Maybe some margin on the bottom?

image

src/css/custom.css
josh-collinsworth: That was definitely a mistake. I was trying to clean up the code before pushing and must have gotten a little over-zealous. 🤦

josh-collinsworth: Oh wait—no, it's just manually set on the body element further up. Still, probably best to be consistent and use the variables they provide.

kwhinnery:
The hover states on the buttons in dark mode might need some love still 

image

kwhinnery (approved):

PR #269, opened 2024-01-12
Let's put that dino in an ellipse

josh-collinsworth:
That dinosaur looks lots better on a black background.

CleanShot 2024-01-12 at 15 46 05@2x


ry (approved):
Iteration is welcome!

PR #282, opened 2024-01-25
Update building_from_source.md

muddlebee:
clang-16 lld-16 is not found in Ubuntu 22.04 which is latest LTS version.
 Similar commands like WSL section works

### DIFF
--- a/runtime/manual/references/contributing/building_from_source.md
+++ b/runtime/manual/references/contributing/building_from_source.md
@@ -48,10 +48,12 @@ cargo -V
 > Many components of Deno require a native compiler to build optimized native
 > functions.
 
-**Linux**:
-
-```sh
-apt install --install-recommends -y clang-16 lld-16 cmake libglib2.0-dev
+**Linus/WSL**
+```shell
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+./llvm.sh 16
+apt install --install-recommends -y cmake libglib2.0-dev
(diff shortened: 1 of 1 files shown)


runtime/manual/references/contributing/building_from_source.md
```diff
@@ -48,10 +48,12 @@ cargo -V
 > Many components of Deno require a native compiler to build optimized native
 > functions.
 
-**Linux**:
-
-```sh
-apt install --install-recommends -y clang-16 lld-16 cmake libglib2.0-dev
+**Linus/WSL**
```
kwhinnery, line 51: `Linus` -> `Linux` here

muddlebee: @kwhinnery its better if we use `Debian` instead of Linux? since apt is supported only in Debian based linux I guess.

muddlebee:
@kwhinnery its better if we use Debian instead of Linux? since apt is supported only in Debian based linux I guess.

runtime/manual/references/contributing/building_from_source.md
kwhinnery: That would probably be ideal, yes

muddlebee: @kwhinnery made the changes like below
Linux -> Linux(Debian)

thisisjofrank (approved):

PR #300, opened 2024-01-30
add more APIs to deprecation guide

iuioiua:
Further work on top of #299. Apologies for the noisy diff - it's due to the re-sorting of sections in alphabetical order of their headings.


iuioiua:
PTAL @kwhinnery

kwhinnery (approved):

PR #302, opened 2024-02-01
fix: syntax highlighting on workers example

tjamesmac:
Syntax highlighting on the [deno-docs site](https://docs.deno.com/runtime/manual/runtime/workers) for these two code blocks was missing. Not sure if the ignores were there on purpose

| Before | After |
| ------- | ----- |
| image    | image   |
|  image | image |

The code blocks seem to work on github before these changes were made


bartlomieju (approved):
Thanks!

PR #303, opened 2024-02-02
Fix the issue of displaying Traditional Chinese when using Google Translate to Simplified Chinese in Mac

jiawei397:
After my testing, I found that using `font-feature-settings: "cv10" on` when translating, it will cause simplified Chinese to be displayed as traditional Chinese. 

image

It would be strange to use software such as Google Translate to translate web pages in full screen. 

Therefore, I have adjusted `cv10` to `cv09`, and I believe it will not have much impact.

### DIFF
--- a/static/fonts/inter.css
+++ b/static/fonts/inter.css
@@ -31,5 +31,5 @@
 }
 
 :root {
-  font-feature-settings: "cv05" on, "cv08" on, "cv10" on, "ss01" on, "ss08" on;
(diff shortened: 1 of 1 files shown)


jiawei397:
This issue has only been found in Chrome and Edge on Mac at the moment. There is no problem with the Safari browser. When set `font-feature-settings` to `cv10`, it will find traditional Chinese fonts in the current font set, and I don't know what's going on exactly.

kwhinnery (approved):

kwhinnery:
thank you!

PR #313, opened 2024-02-04
Explain how to test Deno KV code locally

NoraCodes:
This is easily apparent to users who are already familiar with SQLite, but people who either don't know about ":memory:" or who aren't aware that Deno KV is based on SQLite when run locally may not be able to figure this out for themselves.

I wasn't sure whether to include the assert import for the testing example; if that's needed I can add it.

### DIFF
--- a/deploy/kv/manual/index.mdx
+++ b/deploy/kv/manual/index.mdx
@@ -237,6 +237,45 @@ created by Apple.
 KV on Deploy - a new Deploy database will be provisioned for you when required
 by your code. Learn more about Deno KV on Deno Deploy [here](./on_deploy.mdx).
 
+## Testing
+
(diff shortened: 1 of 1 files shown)


deploy/kv/manual/index.mdx
```diff
@@ -237,6 +237,45 @@ created by Apple.
 KV on Deploy - a new Deploy database will be provisioned for you when required
 by your code. Learn more about Deno KV on Deno Deploy [here](./on_deploy.mdx).
 
+## Testing
+
+By default, [`Deno.openKv()`](https://www.sqlite.org/inmemorydb.html) creates or
```
thisisjofrank, line 242:
```suggestion
By default, [`Deno.openKv()`](https://deno.land/api?unstable=&s=Deno.openKv) creates or
```

thisisjofrank (approved):

PR #318, opened 2024-02-06
Doc: How to delete a deployment

### LINKED ISSUE #286: Document how to delete a deployment
Delete a deployment - help users understand how to manage deployments and that deployments are by default forever. (Like diamonds).

This concept is probably confusing to users that deployments never go away and are accessible forever. Explicitly state that deployments need to be deleted if you need them to go away.

rbetts:
Resolves #286


rbetts:
@kwhinnery - I'd appreciate feedback on how to best reference the dash project page, if you have a point of view?

kwhinnery:
I think our best bet (without a way to deep link to the tab) is to include a screenshot - updated the branch with this technique.

PR #350, opened 2024-02-28
Remove example which depends on the removed std/signal

kt3k:
closes https://github.com/denoland/deno_std/issues/4389


lino-levan (approved):
LGTM.

iuioiua (approved):

PR #376, opened 2024-03-25
update tutorial with new deno deploy flow

thisisjofrank:


magurotuna (approved):
LGTM.
Maybe to avoid confusion, it would be great to mention that you need to answer "No" to the question of whether to use a styling library because tailwind requires AoT build step, though we assume that this tutorial aims to cover the most basic, no-build setting of Fresh. This could probably be further improvement in a follow up.

PR #377, opened 2024-03-25
refactor: replace deno.land/std with jsr:@std

crowlKats:
TODO:
- [ ] update std/examples & testing examples
- [ ] deno.lock looks differently now

Usage of `--eval-file` flag are now broken (https://github.com/denoland/deno/issues/23071) and deployctl doesn't support jsr specifiers for entrypoint


mattvr:
Oh whoops, just PRed almost the same thing and missed this: https://github.com/denoland/deno-docs/pull/587. Mine should be unblocked from landing as-is, but can change its scope or abandon if you prefer. @crowlKats

PR #407, opened 2024-04-08
add docs for check command

thisisjofrank:


runtime/manual/tools/check.md
```diff
@@ -0,0 +1,167 @@
+# deno check
+
+Download and type-check without execution.
```
bartlomieju, line 3:
```suggestion
Download and type-check a program without execution.
```

runtime/manual/tools/check.md
```diff
@@ -0,0 +1,167 @@
...
+
+`deno check [OPTIONS] <FILE>` - Download and type-check `FILE`
+
+## Synopsis
+
+```bash
+deno check [--import-map <FILE>] [--no-remote] [-q|--quiet] [--no-npm] 
+[--node-modules-dir[=<node-modules-dir>]] [--vendor[=<vendor>]]
+[-c|--config <FILE>] [--no-config] [-r|--reload[=<CACHE_BLOCKLIST>...]]
+[--lock [<FILE>]] [--lock-write] [--no-lock] [--cert <FILE>] [--all] <FILE>
```
bartlomieju, line 15: This is useful, but how do we keep it in sync with actual CLI? Eg. we might remove `--node-modules-dir` flag in Deno 2.

thisisjofrank: For now I'm just gonna have to be kept in the loop with changes, my hope for these docs is that they're a little more human friendly than the output from --help is, so we're not quite in a place yet where they could be autogenerated

bartlomieju (approved):

PR #459, opened 2024-05-14
docs: update `deno_std` and import map docs

iuioiua:
Closes https://github.com/denoland/deno_std/issues/4728


kt3k (approved):
It makes sense to me to recommend `jsr:@std` over `deno.land/std`. LGTM

PR #460, opened 2024-05-15
fix: the version number of `/std` and `update_versions.ts`

kt3k:
`getLatestTagForRepo("deno_std")` gets the latest tag from `deno_std` repo, but the latest release tag is not the latest version of `deno.land/std` anymore since the release https://github.com/denoland/deno_std/pull/4689

This PR fixes it.

related https://github.com/denoland/deno_std/issues/4728


dsherret (approved):
LGTM

PR #465, opened 2024-05-17
chore: format source code, run `deno fmt --check` in CI

kt3k:
This PR formats the source code with `deno fmt`, and also sets up the CI task, which checks the code format with `deno fmt --check`.


thisisjofrank (approved):
Nice, lgtm.

Issue #470, opened 2024-05-18
Deno run examples don't seem to work


numbertheory:
Maybe I'm making a very simple mistake, but I've tried this on MacOS and Fedora Linux, both running ZSH shell, with the same results.

I go to the hello world example at this URL:

https://docs.deno.com/examples/hello-world

And copy the text to run it from the footer

```
deno run https://github.com/denoland/deno-docs/blob/main/by-example/hello-world.ts
```
And I get this in my console:

![Screenshot from 2024-05-17 22-08-49](https://github.com/denoland/deno-docs/assets/844696/95007465-073e-43b8-a733-571f2649d510)

Which makes sense because `deno` is trying to run the whole HTML document (github interface and all).

You should probably use this format to provide deno run examples (the raw githubusercontent link).

```
deno run https://raw.githubusercontent.com/denoland/deno-docs/main/by-example/hello-world.ts
```

Which runs as expected:
![Screenshot from 2024-05-17 22-12-51](https://github.com/denoland/deno-docs/assets/844696/1911595a-07d1-40d6-9da3-ca8632cdf3eb)

PR #471, opened 2024-05-18
fix: deno run examples should point to raw githubusercontent domain

numbertheory:
I reported this as issue #470, but it was easy enough to fix, so here's my solution. If there's a better way, let me know, I can update it. Thanks!


thisisjofrank (approved):
Ah good, spot, thank you!

PR #492, opened 2024-06-06
feat: add secondary navbar to docs

### LINKED ISSUE #486: Finalize navigation header on api reference category pages

### LINKED ISSUE #489: Create index pages for node, web, and deno APIs
Example of index page for Deno reference APIs

Image

donjo:
# * Don't merge *

fixes #486 
fixes #489 

Implements a secondary navbar for the upcoming API reference docs section with some sample index pages for the API categories that the reference doc area will contain.

Also updates a number of (largely dark-mode) styles to match the design direction deno.com has gone since this was first built.

### **Todo:**
- [x] Pick a better color to separate the primary and secondary navbar in light mode
- [ ] Update the node and web index pages with content
- [x] Give the index pages proper page headers
- [x] Adjust content width on index pages   

### Preview:

Screenshot 2024-06-10 at 4 22 44 PM


donjo:
@josh-collinsworth ready for review now, just have placeholder content for the API index pages

cc/ @thisisjofrank

src/theme/Layout/index.js
```diff
@@ -18,7 +18,6 @@ import styles from "./styles.module.css";
 export default function Layout(props) {
   const {
     children,
-    noFooter,
     wrapperClassName,
```
josh-collinsworth, line 21: I don't know if it's related to this or not, but it looks like the footer is jumping up and overlapping content. 

image
image

josh-collinsworth (changes requested):
General question: when will we merge this? Will Leo build on this PR to flesh out the content?

josh-collinsworth:
Forgot to note this in the review, but just for posterity: we'll want to account for this overlap wherever that subnav is used

src/theme/Layout/index.js
donjo: I think this was from a lingering h-screen on the content container which I removed after you probably already started reviewing so see if you can still reproduce as it may be fixed.

tailwind.config.js
```diff
@@ -7,7 +7,56 @@ module.exports = {
...
+
+        fresh: "#FFDB1E",
+        "fresh-dark": "#401C00",
+
+        gray: {
+          0: "#9EA0A5",
+          1: "#868789",
+          2: "#56575A",
+          3: "#25272B",
+          4: "#191B1F",
```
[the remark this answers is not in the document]
donjo: I had to put quotes around them to not throw linting errors, does that sound right?

src/theme/Layout/index.js
josh-collinsworth: Yeah I think that must've been it. 👍

donjo:
> General question: when will we merge this? Will Leo build on this PR to flesh out the content?

@josh-collinsworth 

Leo is going to build the nav in the app he's building to serve the docs content so the nav may not actually live in docusaurus at this time but I may keep it in this PR as an un-used component we could use elsewhere in the docs if we wanted and as a reference for what the styles should look like given the globally set docusaurus variables so I'm going to clean up this PR a bit given that strategy and make the API Reference link link to the external link it goes to now until we're fully ready to join these two apps.

donjo:
closing as this was covered in #519

PR #510, opened 2024-06-22
fix(runtime/manual/tools/init): remove bench

sntran:
`deno init` no longer creates the bench file.
Output also does not include the instruction to run benchmarks.


ry:
thanks

PR #524, opened 2024-07-03
fix: remove hardcoded html from markdown table

### LINKED ISSUE #522: HTML code not rendered in TypeScript section
https://docs.deno.com/runtime/manual/advanced/typescript/configuration/#how-deno-uses-a-configuration-file

Screenshot 2024-07-03 at 16 58 14

donjo:
fixes #522

### DIFF
--- a/runtime/manual/advanced/typescript/configuration.md
+++ b/runtime/manual/advanced/typescript/configuration.md
@@ -41,7 +41,7 @@ ignored.
 Here is a table of compiler options that can be changed, their default in Deno
 and any other notes about that option:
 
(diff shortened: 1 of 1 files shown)


ry (approved):

thisisjofrank:
wait, this was added on purpose to fix the alignment of the cells, has that been fixed elsewhere now?

donjo:
@thisisjofrank do you recall what the alignment looked like before in Docusaurus? I had figured those were hard coded widths to make sure the text in the `option` and `default` columns weren't wrapping to a new line before the string was complete. The migration to Lume (and however it parses markdown I guess) caused the manual styles to not actually go into effect but just render as text in the header. I removed them to clean that up and the table below seemed to render OK but if there's a different output we're looking for we can do that, it just has to be done differently than it was with Docusaurus.

PR #530, opened 2024-07-04
adding docs on private registries

thisisjofrank:


runtime/manual/basics/modules/grpc_connections.md
```diff
@@ -0,0 +1,25 @@
+---
+title: "gRPC Connections"
+---
+
+gRPC is a high-performance, open-source, universal RPC framework that enables
+efficient communication between services. With gRPC support, you can build
```
bartlomieju, line 6: What prompted this one? GRPC is hardly in the "basics" field and I think we shouldn't mention it so prominently in the manual 😅

runtime/manual/basics/modules/private_registries.md
```diff
@@ -0,0 +1,63 @@
+---
+title: "Private Registries"
```
bartlomieju, line 2: To make it clearer that we're talking about npm, can we rename this page to something like `npm private registries`? Also, does it make sense to be in the "basic" page and not in the "Node compat" section?

runtime/manual/basics/modules/private_registries.md
```diff
@@ -0,0 +1,63 @@
...
+Large organizations often host their own private npm registries to manage
+internal packages securely. These private registries serve as repositories where
+organizations can publish and store their proprietary or custom packages. Unlike
+public npm registries, private registries are accessible only to authorized
+users within the organization.
+
+## How to use private registries with Deno
+
+First, configure your `.npmrc` file to point to your private registry. You can
+do this by adding the following line to your `.npmrc` file:
```
bartlomieju, line 20: Maybe worth linking to npm's own docs here as well?

runtime/manual/basics/modules/private_registries.md
bartlomieju, line 1: Suggest to mention the limitation that we only read `.npmrc` in the project root and `$HOME` directory; and these files are not yet combined: https://github.com/denoland/deno/issues/23954

bartlomieju (approved):

runtime/manual/basics/modules/grpc_connections.md
thisisjofrank: it was in the 1.4 release blog and didn't have documentation, so thought I'd add it in. Is there a better place for it to go?

runtime/manual/basics/modules/private_registries.md
thisisjofrank: Which ones were you thinking particularly?

runtime/manual/basics/modules/private_registries.md
thisisjofrank: Just to double check that means the file must be in both places or its an either or?

runtime/manual/basics/modules/grpc_connections.md
ry: I think this at least belongs in a different PR but I agree with Bartek that maybe not in the manual at all. The support Deno provides is better node compatibility to run gRPC packages - it doesn’t have direct gRPC support.

runtime/manual/basics/modules/private_registries.md
bartlomieju: We could link to https://docs.npmjs.com/cli/v10/configuring-npm/npmrc

runtime/manual/basics/modules/grpc_connections.md
thisisjofrank: I'll remove it

runtime/manual/basics/modules/private_registries.md
bartlomieju: Currently it's either, which is not aligned with how `npm` works

PR #555, opened 2024-07-08
Restyle admonition blocks

donjo:
Screenshot 2024-07-08 at 3 22 35 PM

Restyled these blocks to stand out a bit more and feel more unique between each admonition type


crowlKats:
note and info are so hard to distinguish, bothers me that they have different shades, might as well have them be the same color since they have the same icon?

crowlKats (approved):

PR #559, opened 2024-07-08
chore: add hint for mapping internal modules

marvinhagemeister:
Didn't know about the ability to map internal modules. This adds the missing hint to the docs.


iuioiua (approved):
Nice

PR #566, opened 2024-07-10
Fix image paths

zebreus:
Fixes the image paths on https://docs.deno.com/runtime/manual/advanced/private_repositories/


donjo (approved):

donjo:
thanks!

PR #569, opened 2024-07-11
Revert "chore: update for 1.45.0"

donjo:
Reverts denoland/deno-docs#568 because the mobile navigation doesn't load on 1.45.


bartlomieju (approved):

donjo:
not needed anymore, figured out the problem

PR #571, opened 2024-07-11
fix: remove upstreamed ddoc stylings and update README

crowlKats:
moved stylings that are deno-docs specific inton the styles.css, and things still in the overrides css file are still to be backported.
deno_doc PR denoland/deno_doc#611


overrides.css
```diff
@@ -1,149 +1,9 @@
...
-  @apply flex-grow;
-}
-
-.toc {
-  @apply hidden;
-}
-
-.ddoc > .toc {
-  @apply hidden md:flex md:flex-col !important;
-}
```
donjo, line 23: The styles here that adjust the ToC position are still necessary because if all of ddoc is in a grid it doesn't anchor the navigation content to the left with a fixed width so I think ddoc should be a flex where the categories list is on the left and the rest of the content takes up the rest of the space. 

We could either keep that in here until the next deno_doc pass or migrate it to styles.css as something that is a deno-docs specific interpretation

Screenshot 2024-07-11 at 9 05 39 AM

overrides.css
```diff
@@ -1,149 +1,9 @@
...
-  > div
-  > div#content
-  > div.toc
-  > div
-  > nav.documentNavigation
-  > ul
-  > li
-  > ul
-  > li {
-  @apply mt-1 !important;
```
donjo, line 122: Some of these long document navigation styles should move or else we lose the formatting of the document navigation

Screenshot 2024-07-11 at 9 51 30 AM
Screenshot 2024-07-11 at 9 49 19 AM

donjo (changes requested):
Screenshot 2024-07-11 at 9 49 19 AM

donjo (approved):

donjo:
whoops forgot to run `deno task doc` so disregard the problems I found

PR #584, opened 2024-07-12
Delete static/openapi.json

piscisaureus:


thisisjofrank (approved):
Seems legit, doesn't seem to be referenced anywhere in the docs! I wonder why it was there.

PR #586, opened 2024-07-12
truncating navigation buttons

donjo:
Screenshot 2024-07-12 at 2 25 13 PM

This ensures that the navigation text on the doc nav buttons truncates so one button doesn't have more height than another.

### DIFF
--- a/_includes/doc.tsx
+++ b/_includes/doc.tsx
@@ -214,15 +214,13 @@ function NavigationButton(
       href={"id" in item ? item.id : "href" in item ? item.href : undefined}
     >
       <span className="text-sm text-gray-2">{directionText}</span>
-      <span
-        className={`font-semibold text-blue-500 ${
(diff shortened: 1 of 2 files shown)


_includes/doc.tsx
```diff
@@ -139,24 +139,20 @@ export default function Page(props: Lume.Data, helpers: Lume.Helpers) {
...
                 </div>
                 <div>
                   {parentNavigation[index! + 1] && (
-                    <div>
-                      <NavigationButton
-                        item={parentNavigation[index! + 1]}
-                        search={props.search}
-                        direction="next"
-                      />
-                    </div>
```
donjo, line 159: removing what seemed like an unnecessary extra div here

donjo, line 159: removing what seemed like an unnecessary div here

PR #597, opened 2024-07-15
Navigation Updates

donjo:
This PR updates a number of nav styles on both the top nav and side nav

## Top Nav
- Content is re-arranged to group runtime and deploy content
- Style changes to fix alignment at different mobile viewports
- More subtle active state accent underlines
- Search box resizing 
- Slight spacing adjustments around logo

## Table of contents (API Ref)
- Removed header since there is only one section of content and every item is in that section
- Spacing changes to adjust for that.
- ToC width reduced since the table of contents only provides one level of navigation with no subnav content and thus don't utilize the arrow > icon like the rest of the docs do.

## Other
- Body content spacing adjustments to account for changes to ToC

Screenshot 2024-07-15 at 4 29 14 PM

### DIFF
--- a/_components/Header.tsx
+++ b/_components/Header.tsx
@@ -12,13 +12,13 @@ export default function Header({
         reference ? "" : "sticky top-0 left-0 right-0"
       }`}
     >
-      <nav class="px-4 md:px-6 py-3 h-16 flex items-center justify-between">
+      <nav class="px-4 md:px-6 pt-2.5 pb-2 h-16 flex items-center justify-between">
         <div class="flex items-center">
           {hasSidebar && (
             <button class="mr-2 lg:hidden" id="sidebar-open">
               <svg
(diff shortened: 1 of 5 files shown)


_components/Header.tsx
```diff
@@ -33,57 +33,62 @@ export default function Header({
...
+            <span style={{ fontSize: "1.375rem" }} class="font-semibold">
+              Docs
+            </span>
           </a>
           <HeaderItem
             url={url}
             activeOn="/runtime"
             href="/runtime/manual"
-            name="Deno Runtime"
+            name="Runtime"
```
ry, line 49: "Runtime" or "Manual"? "Runtime Manual" ?

donjo: I like Runtime Manual a bit better I think

PR #598, opened 2024-07-16
Update path-operations example

kt3k:
The current example uses outdated `path.win32` and `path.posix` objects, which were deleted in https://github.com/denoland/deno_std/pull/4953. This PR updates them to the appropriate alternatives.


thisisjofrank (approved):
Lgtm

PR #610, opened 2024-07-18
fix(server): explicitly set content-type to text/html for requests to /api

### LINKED ISSUE #607: Navigating to reference for Deno.run downloads page HTML
Going to https://docs.deno.com/api/deno/~/Deno.run triggers a download of the page's HTML instead of displaying it.

I think this may be due to a wrong `Content-Type` header; `curl -I https://docs.deno.com/api/deno/~/Deno.run` says this;
```http
HTTP/2 200
accept-ranges: bytes
content-length: 34291
content-type: application/x-makeself
// ^ here
etag: W/"c-wLIEy3j86nF4fjXU0zU59UNrUah"
server: deno/gcp-europe-west3
vary: Accept-Encoding
date: Wed, 17 Jul 2024 21:03:39 GMT
via: http/2 edgeproxy-h
```

In comparison, a page that loads normally (Deno.CommandOutput) shows this:
```http
HTTP/2 200
accept-ranges: bytes
content-length: 19894
etag: W/"c-wLIEy3j86nF4fjXU0zU59UNrUah"
server: deno/gcp-europe-west3
vary: Accept-Encoding
date: Wed, 17 Jul 2024 21:04:01 GMT
via: http/2 edgeproxy-h
```
I don't see a `Content-Type` header here for some reason, which is why I'm not 100% sure of this being the cause.

I'm using Firefox 128 on Windows 11.

### LINKED ISSUE #609: API doc of `Deno.Kv.prototype.list` is rendered as plain text
https://docs.deno.com/api/deno/~/Deno.Kv.prototype.list

Opening this URL with whatever browser will end up with plain text. 

open Deno.Kv.prototype.list in chrome

magurotuna:
This commit adds a custom middleware to overwrite content-type in the response header with `text/html` for all requests to `/api` except for CSS and JS.
This helps avoid the MIME type detection errors that happen with some file names such as `Deno.Kv.prototype.list`.

Fixes #609 
Fixes #607

---

With the local dev server I confirmed that Deno.Kv.prototype.list is rendered correctly and content-type in the response header is set to text/html.

Deno.Kv.prototype.list works correctly on local


deno.lock
```diff
@@ -87,9 +87,11 @@
       "npm:mongodb@6.1.0": "npm:mongodb@6.1.0",
       "npm:postcss-import@16.1.0": "npm:postcss-import@16.1.0_postcss@8.4.38",
       "npm:postcss@8.4.38": "npm:postcss@8.4.38",
+      "npm:postcss@8.4.39": "npm:postcss@8.4.39",
       "npm:preact": "npm:preact@10.22.0",
       "npm:preact-render-to-string@6.5.5": "npm:preact-render-to-string@6.5.5_preact@10.22.0",
       "npm:preact@10.22.0": "npm:preact@10.22.0",
+      "npm:preact@10.22.1": "npm:preact@10.22.1",
```
marvinhagemeister, line 94: Let's stay on `10.22.1`, need to release a new Fresh 1.x version to make Fresh work with latest Preact

middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+  // We need to explicitly set the content type here because some file names
+  // confuse the MIME type detection - for example, the file name
+  // `Deno.Kv.prototype.list` is inferred as `text/plain` but it should be
+  // `text/html` for browsers to render it correctly.
+  const requestToApiDoc = new URL(request.url).pathname.startsWith("/api");
+  const contentTypeSetToCssOrJs = isCssOrJs(
+    response.headers.get("content-type"),
+  );
+  if (requestToApiDoc && !contentTypeSetToCssOrJs) {
+    response.headers.set("content-type", "text/html");
```
magurotuna, line 21: note to self: let's specify character encoding

crowlKats (approved):
LGTM

PR #613, opened 2024-07-19
doc: Add instructions for building from source

irbull:
Adds instructions for building from source when the change-set spans multiple crates.

### DIFF
--- a/runtime/manual/references/contributing/building_from_source.md
+++ b/runtime/manual/references/contributing/building_from_source.md
@@ -171,3 +171,42 @@ cargo clean && cargo build -vv
 # Run:
 ./target/debug/deno run tests/testdata/run/002_hello.ts
 ```
+
+## Working with Multiple Crates
+
+If a change-set spans multiple Deno crates, you may want to build multiple
+crates together. It's suggested that you checkout all the required crates next
+to one another. For example:
+
+```shell
+- denoland/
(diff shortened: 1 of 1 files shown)


runtime/manual/references/contributing/building_from_source.md
```diff
@@ -171,3 +171,28 @@ cargo clean && cargo build -vv
...
+  - ...
+```
+
+Then you can use
+[Cargo's patch feature](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html)
+to override the default dependency paths:
+
+```shell
+cargo build --config 'patch.crates-io.deno_ast.path="../deno_ast"'
+```
```
bartlomieju, line 195: This might get cumbersome if you are working on something for several days - maybe it's worth to mention another option which is changing `Cargo.toml` by adding these lines:

```
[patch.crates-io]
deno_ast = { path = "../deno_ast" }
```

irbull: Sounds good, although my auto complete in fish makes this so easy :).

Yes, I'm happy to add this too. Obviously they need to make sure they don't stage this change. Another thing to mention is that the version numbers must still match.

I'll update the PR.

bartlomieju (approved):
LGTM, thanks!

PR #632, opened 2024-07-25
switch APIs to single col layout

donjo:
Screenshot 2024-07-25 at 12 06 53 PM

This adds style overrides to render the APIs in a single column layout with some adjusted spacing to account for the new layout.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -272,4 +272,7 @@ body:not(:has(.ddoc)) {
       background-color: var(--color-neutral-muted) !important;
     }
   }
+  .markdown_summary :not(pre) > code {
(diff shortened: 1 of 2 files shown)


styles.css
```diff
@@ -272,4 +272,7 @@ body:not(:has(.ddoc)) {
       background-color: var(--color-neutral-muted) !important;
     }
   }
+  .markdown_summary :not(pre) > code {
+    background-color: var(--color-neutral-muted) !important;
```
donjo, line 276: this is unrelated to the single col thing I just noticed that the style above didn't catch all code blocks

crowlKats (approved):

PR #635, opened 2024-07-26
fix: add missing cli subcommands and update descriptions

rbetts:
Compared this content to `deno -h` and updated accordingly. Note that I linked `add` and `publish` to docs in other parts of the manual tree. On the fence if that's lazy or clever - but the linked sections contain useful content that seemed unnecessary to duplicate.


thisisjofrank (approved):
We should definitely give them their own pages, as the other commands, but this works as an interim for sure.

Issue #654, opened 2024-08-05
`deno-dom` manual page links to  `deno.land`


marvinhagemeister:
`deno-dom` manual page links to deno.land https://docs.deno.com/runtime/manual/advanced/jsx_dom/deno_dom/ instead of JSR

PR #655, opened 2024-08-05
fix(deno_dom): point to JSR isntead of deno.land

marvinhagemeister:
Fixes https://github.com/denoland/docs/issues/654


kt3k (approved):
Nice!

PR #682, opened 2024-08-06
wip: create a style guide for error messages

irbull:
Creates an initial style guide for user-facing error messages. The guide is based on the style guide from V8.

Two large questions that need to be answered (by a core team member)

1. What should the styles be? These are just suggestions / placeholders, but a core team member should really decide on the styles.
2. Will the existing code-base be updated to conform to a consistent set of styles? This doesn't need to happen all at once, but if there is no appetite to provide consistency across the code-base, then the style guide will provide more harm than good.

https://github.com/denoland/std/issues/5574

### DIFF
--- a/runtime/manual/references/contributing/style_guide.md
+++ b/runtime/manual/references/contributing/style_guide.md
@@ -356,6 +356,75 @@ export function foo(): string {
 }
 ```
 
+### Error Messages
+
+User-facing error messages should be clear, concise, and consistent. Error
+messages should be in sentence case but should not end with a period. Error
+messages should be free of grammatical errors and typos and written in American
+English.
+
+> ⚠️ Note that the error message style guide is a work in progress, and not all
+> the error messages have been updated to conform to the current styles.
+
+Error message styles that should be followed:
+
+1. Messages should start with an upper case:
+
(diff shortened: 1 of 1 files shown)


iuioiua:
I'll close this as we should start with `std` first. Please open an issue there 🙂

irbull:
Just to check, are you looking for a documentation change to `std`, or would you like help with actually formatting the messages? I'm happy to move this there, but I think having consistent messages between the `std` and anything produced by the runtime would also make sense. Also, giving contributors guidelines for error messages, is probably valuable (but maybe after the `std` / runtime are updated).

iuioiua:
Yes, we're happy to have a documentation change for `std`. That can be done in the contributing guidelines in that repo. It's best that we play with the idea in `std` before enforcing it in the runtime and all other repos.

irbull:
I've filed a PR in the std: https://github.com/denoland/std/pull/5671

I'll be off-line without net access for the next week. Feel free to do with this what you want, otherwise I can revisit it when I return.

iuioiua:
Great stuff! Thank you. I'll look at it this week (I was away last week).

PR #688, opened 2024-08-07
Adjust sidenav and details colors

donjo:
Screenshot 2024-08-07 at 9 36 24 AM

Screenshot 2024-08-07 at 9 36 54 AM

Two instances of removing light pink from our UI since we've settled into using blue as our primary color until the new dotcom is done.  

Also fixed some nearby minor spacing issues with the docs content body where content was too close to the sidenav at the largest viewport


josh-collinsworth (approved):
Da bo dee da bo di

PR #697, opened 2024-08-09
feat: highlight active heading in sticky toc

marvinhagemeister:
Enhances the table of contents component on the right to show the active headline

https://github.com/user-attachments/assets/fe81e186-9201-45db-a719-1a393b08bcc8

### DIFF
--- a/_includes/layout.tsx
+++ b/_includes/layout.tsx
@@ -2,7 +2,7 @@ export default function Layout(props: Lume.Data) {
   const reference = props.url.startsWith("/api");
 
   return (
-    <html lang="en" class={reference ? "" : "h-dvh overflow-hidden"}>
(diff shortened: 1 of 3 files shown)


_includes/layout.tsx
```diff
@@ -2,7 +2,7 @@ export default function Layout(props: Lume.Data) {
   const reference = props.url.startsWith("/api");
 
   return (
-    <html lang="en" class={reference ? "" : "h-dvh overflow-hidden"}>
```
marvinhagemeister, line 5: Overflow on parents prevents child elements from being able to use `position: sticky`. When we encounter an overflow issue we should address that instead of using a sledge hammer on a parent element.

josh-collinsworth:
I think this can probably be closed now; thanks for letting me ape your JS for #699 !

PR #699, opened 2024-08-09
Responsive/a11y improvements

josh-collinsworth:
- Fixes issue where `->` arrows on the homepage would sometimes break, and improves link text wrapping
- Makes doc sidebar sticky
  - **NOTE**: this required removing `overflow-x-hidden` from the `<main>` element. In my limited testing, this didn't cause any issues, but thorough reviews appreciated
- Moves the right sidebar out of `<main>` and into an `<aside>` element
- Minor CSS adjustments that mostly shouldn't change anything visually but make things a little easier to reason about under the hood (change `flex` to `grid`).

image

CleanShot 2024-08-09 at 10 07 17

### DIFF
--- a/index.page.tsx
+++ b/index.page.tsx
@@ -233,7 +233,8 @@ export default function () {
                 </h3>
                 <p class="max-w-[75ch]">
                   Deno, Web, and Node API reference documentation, built for the
-                  Deno experience. Explore the APIs available in Deno in the{" "}
+                  Deno experience. Explore the APIs available in Deno in the
+                  {" "}
                   <a
                     href="/api/deno"
                     class="runtime-link underline underline-offset-4"
@@ -457,8 +458,9 @@ function DocsCTA(props: {
   href: string;
   product: "deploy" | "runtime";
 }) {
-  const productClass =
-    props.product === "deploy" ? "deploy-cta" : "runtime-cta";
+  const productClass = props.product === "deploy"
+    ? "deploy-cta"
+    : "runtime-cta";
   return (
     <a href={props.href} className={`docs-cta ${productClass}`}>
       {props.text}{" "}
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):

index.page.tsx
```diff
@@ -487,27 +487,22 @@ function ContentItem(
...
-      <a
-        className={`homepage-link ${productClass}`}
-        href={props.link}
-      >
-        {props.linktext} <span aria-hidden="true">-&gt;</span>
+      <p className="mb-3">{props.description}</p>
+      <a className={`homepage-link ${productClass}`} href={props.link}>
+        {props.linktext}{" "}
+        <span aria-hidden="true" class="whitespace-pre">
+          -&gt;
```
ry, line 494: why not use a unicode arrow character here? EG ➞

josh-collinsworth: The arrow ligature is actually part of the Inter typeface, so it goes with the text, wheres a unicode arrow will unavoidably look unicode-y. Here's an example (the right arrow is the unicode arrow):

CleanShot 2024-08-09 at 13 20 02@2x

styles.css
```diff
@@ -164,6 +164,14 @@ body:not(:has(.ddoc)) {
   }
 }
 
+.sidebar__current-item {
+  @apply text-blue-500 relative;
+}
+
+.sidebar__current-item::before {
+  @apply content-[""] absolute -left-[0.85em] w-[0.5em] text-center h-[0.7em] top-[0.25em] [clip-path:polygon(0%_0%,100%_50%,0%_100%)] bg-blue-500;
```
donjo, line 172: I think given the secondary indicator here we should bump up the padding on the `ul` to pl-4 instead of pl-2 and then make this `-left-[1em]` to visually center it in the new whitespace otherwise it felt a bit crowded to me.

donjo: Screenshot 2024-08-09 at 2 48 17 PM

josh-collinsworth: For posterity: we decided on a different direction for the "current" indicator arrow

PR #702, opened 2024-08-12
feat: add custom tailwind plugin for perf

marvinhagemeister:
Instead of processing all generated files, it processes the input files only. These are far fewer and thus much faster.

In total this reduces tailwind processing time from 2.6s to ~70-80ms on my machine.


thisisjofrank (approved):
This quite considerably speeds things up for me on windows 👍

PR #703, opened 2024-08-12
Exampleurls

thisisjofrank:
updating by-example examples to use a hosted version of the file instead of github raw to make urls a bit prettier


lucacasonato (approved):
LGTM, thank you!

PR #709, opened 2024-08-15
Add overflow scroll to table of contents sidebar

austincrim:
If the table of contents is sufficiently long, it's not possible to view it all due to its sticky position. This PR adds the same overflow treatment as the left sidebar.

Before:

https://github.com/user-attachments/assets/baf79464-1b23-400a-bd5f-4a24f1fed59d

After:

https://github.com/user-attachments/assets/07b5694f-5fd1-4c3b-852e-67a1d3a5185b


_includes/doc.tsx
```diff
@@ -168,7 +168,11 @@ export default function Page(props: Lume.Data, helpers: Lume.Helpers) {
           </div>
         </main>
         <aside class="hidden lg:block pb-8 pr-8 col-span-2">
-          <div class="py-2 sticky top-4" id="toc">
+          <div
+            class="py-2 sticky overflow-y-auto top-4 h-[calc(100vh-7rem)]"
+            style={{ scrollbarGutter: "stable", scrollbarWidth: "thin" }}
```
donjo, line 173: Thanks for your contribution!

Would you mind moving the `style={{` as a style for `#toc` at the bottom of styles.css? I want to avoid the in-line styles (other than tailwind classes) if we can.

donjo (changes requested):

_includes/doc.tsx
austincrim: Sure thing!

donjo (approved):

PR #720, opened 2024-08-19
adding permissions page

thisisjofrank:
should there be an introduction to the -A flag in here?

### DIFF
--- a/runtime/fundamentals/permissions.md
+++ b/runtime/fundamentals/permissions.md
@@ -1,0 +1,308 @@
+---
+title: "Permissions"
+oldUrl:
+  - /runtime/manual/basics/permissionsDeno/
+  - /manual/basics/permissions
+  - /runtime/manual/basics/permissions
+---
+
+Deno is secure by default. Therefore, unless you specifically enable it, a
+program run with Deno has no access to sensitive APIs, such as file system access, network connectivity, or environment access. You must explicitly grant access to these resources with command line flags or with the runtime permission prompt. This is a major difference from Node, where
+dependencies are automatically granted full access to everything, potentially
+introducing hidden vulnerabilities into your project.
+
+## Granting permissions
+
+To grant a permission to a script, you can use the `--allow-<PERMISSION>` flag
+when running the script. For example, to grant read access to the file system,
+you can use the `--allow-read` flag:
+
+```shell
+deno run --allow-read mod.ts
+```
+
+`mod.ts` has been granted read-only access to the file system. It cannot write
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/permissions.md
```diff
@@ -3,33 +3,39 @@ title: "Permissions"
...
 ---
 
 Deno is secure by default. Therefore, unless you specifically enable it, a
-program run with Deno has no file, network, or environment access. Access to
-security sensitive functionality requires that permissions have been granted to
-an executing script through command line flags, or a runtime permission prompt.
-This is a major difference from Node, where dependencies are automatically
-granting full access to everything, introducing hidden vulnerabilities in your
-project.
+program run with Deno has no access to sensitive APIs, such as file system access, network connectivity, or environment access. You must explicitly grant access to these resources with command line flags or with the runtime permission prompt. This is a major difference from Node, where
```
ry, line 10: apply `deno fmt`

runtime/fundamentals/permissions.md
```diff
@@ -3,33 +3,39 @@ title: "Permissions"
...
 
 ```shell
 deno run --allow-read mod.ts
 ```
 
+`mod.ts` has been granted read-only access to the file system. It cannot write
+to the file system, or perform any other security sensitive functions. To deny
+access to a permission, you can use the `--deny-<PERMISSION>` flag. For more
+examples of what you can do with different permissions, check out
+[Deno by Example](https://docs.deno.com/examples/).
```
ry, line 28: deny is the default - this reads a bit weird - why would you want a deny flag if it's the default? the reason is that you can narrow the scope of the permissions. For example, maybe you want to do something like this
```
--allow-read=/Users --deny-read=/Users/baduser
```
I think this deserves its own section and be better explained.

runtime/fundamentals/permissions.md
```diff
@@ -3,33 +3,39 @@ title: "Permissions"
...
-securely and host this code through
-[Deno Subhosting](https://deno.com/subhosting) or any other cloud platform of
-your choice.
-
-For the following example `mod.ts` has been granted read-only access to the file
-system. It cannot write to the file system, or perform any other security
-sensitive functions.
+To grant a permission to a script, you can use the `--allow-<PERMISSION>` flag
+when running the script. For example, to grant read access to the file system,
+you can use the `--allow-read` flag:
```
ry, line 18: You can also discuss the short versions of these flags, which are added in deno 1.46 - `-R` for `--allow-read` etc

marvinhagemeister (approved):

PR #721, opened 2024-08-19
Fix examples sort order

josh-collinsworth:
Fixes the sort order for Examples categories. Also adds some VS code autosave settings so personal "format on save" preferences don't overwrite the docs settings. (As a consequence, there are several `fmt` changes lumped in here, but the sort order thing is the only actual user-facing code change.)

### DIFF
--- a/examples.page.tsx
+++ b/examples.page.tsx
@@ -51,10 +51,9 @@ export default function* (_data: Lume.Data, helpers: Lume.Helpers) {
   });
 
   for (const example of examples) {
-    const contentNoCommentary = example.parsed.files
-      .map((file) => file.snippets.map((snippet) => snippet.code).join("\n"))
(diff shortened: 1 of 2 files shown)


examples.page.tsx
```diff
@@ -215,7 +222,7 @@ export default function* (_data: Lume.Data, helpers: Lume.Helpers) {
       return null;
     }
 
-    group.sort((a, b) => a.sortOrder - b.sortOrder);
+    group.sort((a, b) => a.parsed.sortOrder - b.parsed.sortOrder);
```
josh-collinsworth, line 225: This is the only actual code change in this PR; the rest is just `fmt` changes.

thisisjofrank (approved):
gyat

PR #722, opened 2024-08-19
Typographic and layout fixes

josh-collinsworth:
- Fixes Inter font not displaying properly, especially on Windows (specificity issue with GFM.css)
- Adds some safety for nested CSS (adds `&` for nested element selectors) because of iOS
- Various accessibility upgrades for the sidebar
  - Set heading level to proper `h2` level instead of `h3` (heading order is independent of main document in this case)
  - Make clickable "dropdown" divs into proper `button` elements instead
  - Set `aria-labelledby` on lists as an enhancement
  - Put title in text instead of mixing image and text
- Change sidebar headings to standard case (no longer all uppercase)
- Remove "external link" icon from footer
- Fix issue with scrollbar on sidebar not going all the way to the bottom of the page
- Minor typographic updates


_includes/raw.tsx
```diff
@@ -22,10 +22,15 @@ export default function Raw(props: Lume.Data, helpers: Lume.Helpers) {
...
             <div class="p-4 shadow-sm flex justify-between h-16">
               <a class="flex items-center gap-3 mr-6" href="/">
-                <div class="block size-6">
-                  <img src="/img/logo.svg" alt="Deno Docs" />
-                </div>
-                <b class="text-xl">Docs</b>
+                <img
+                  src="/img/logo.svg"
+                  class="block size-6"
+                  alt=""
```
thisisjofrank, line 28: how come emptied the Alt here?

josh-collinsworth: The old way, ATs would read this "Deno docs docs" (i.e., "docs" twice, since the word was in both the alt and the following text).

This way, ATs ignore the image, which doesn't really provide any additional context, and simply announces "Deno docs." The text is all in HTML as well, which is better for i18n.

donjo (approved):

PR #726, opened 2024-08-20
Moving standard library file

thisisjofrank:


runtime/fundamentals/standard_library.md
```diff
@@ -0,0 +1,82 @@
...
+```ts
+import { copy } from "@std/fs";
+import { join } from "@std/path";
+
+await copy("foo.txt", join("dist", "foo.txt"));
+```
+
+Alternatively, you can import modules directly with the `jsr:` specifier:
+
+```json
```
thisisjofrank, line 52:
```suggestion
```js
```

runtime/fundamentals/standard_library.md
```diff
@@ -0,0 +1,82 @@
...
+import { copy } from "@std/fs";
+import { join } from "@std/path";
+
+await copy("foo.txt", join("dist", "foo.txt"));
+```
+
+Alternatively, you can import modules directly with the `jsr:` specifier:
+
+```js
+// This works without import map
```
thisisjofrank, line 53:
```suggestion
```

runtime/_data.ts
```diff
@@ -18,9 +18,11 @@ export const sidebar = [
     items: [
       "/runtime/fundamentals/ts_support/",
       "/runtime/fundamentals/permissions/",
+      "/runtime/fundamentals/standard_library/"
       "/runtime/fundamentals/debugging/",
       "/runtime/fundamentals/workspaces/",
       "/runtime/fundamentals/stability/",
+
```
marvinhagemeister, line 25: nit space, maybe a `deno fmt` was missing.

marvinhagemeister (approved):

runtime/_data.ts
thisisjofrank, line 25:
```suggestion
```

PR #727, opened 2024-08-20
Joco/copy btn fix

josh-collinsworth:
Looks like this issue was already fixed, but the `hidden` class is kind of problematic anyway, since it removes the element from the DOM entirely, so now we just switch its opacity instead.

### DIFF
--- a/copy.client.ts
+++ b/copy.client.ts
@@ -3,13 +3,13 @@ function findParent(el, find) {
     if (find(el)) {
       return el;
     }
-  } while (el = el.parentElement);
+  } while ((el = el.parentElement));
 }
 
 document.addEventListener("click", (e) => {
   const target = findParent(
     e.target,
-    (el) => el instanceof HTMLButtonElement && el.dataset["copy"],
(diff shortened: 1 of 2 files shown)


copy.client.ts
```diff
@@ -3,7 +3,7 @@ function findParent(el, find) {
     if (find(el)) {
       return el;
     }
-  } while (el = el.parentElement);
+  } while ((el = el.parentElement));
```
thisisjofrank, line 6: is this some weird formatting rule?

josh-collinsworth: Now that I'm looking at this...I'm wondering if we should actually just rewrite this whole file. Seems odd to add the listener to the whole document and do this much DOM crawling. (I wasn't concerned with the JS at the time; I was just working on the CSS.) What do you think @thisisjofrank?

josh-collinsworth: Never mind; I went ahead and pulled the trigger on a rewrite, which I think makes much more sense (and is probably more performant) anyway. I also modified the styling a little bit.

thisisjofrank (approved):
a million times better

PR #731, opened 2024-08-21
adding testing page

thisisjofrank:
this is a long one, sorry!


runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,739 @@
...
+// test with permissions
+Deno.test({
+  name: "read file test",
+  permissions: { read: true },
+  fn: () => {
+    const data = Deno.readTextFileSync("./somefile.txt");
+    assertEquals(data, "expected content");
+  },
+});
+```
```
marvinhagemeister, line 62: Nit: we should maybe hint that there is also a jest-like `expect` API in std via https://jsr.io/@std/expect for folks who prefer that. Let's add one example which uses that.

marvinhagemeister (approved):

runtime/fundamentals/testing.md
thisisjofrank: added <3

PR #743, opened 2024-08-23
docs: add "linting and formatting" page

thisisjofrank:


runtime/fundamentals/linting_and_formatting.md
```diff
@@ -0,0 +1,86 @@
...
+
+By default, `deno fmt` formats all TypeScript and JavaScript files in the
+current directory and its subdirectories. If you want to format specific files
+or directories, you can pass them as arguments to the command. For example:
+
+```bash
+deno fmt src/
+```
+
+This command will format all files in the `src/` directory.
```
marvinhagemeister, line 67: I wonder if we should add a note about `deno fmt --check` to check formatting on CI. That's likely the thing people want to know about to ensure consistent formatting. Or maybe we should have something in this doc that points to how to set up Deno on CI.

thisisjofrank: I've added in an extra section

marvinhagemeister:
Whoops I missed the todo links, will do a follow up PR

PR #746, opened 2024-08-23
Docs refresh

thisisjofrank:
Create new navigation fields with a view to flattening the docs structure

- Runtime landing
- Getting Started
    - Hello World
    - Set up your environment
    - Command Line Interface
- Fundamentals
    - TS Support
    - Security
    - Modules ( not yet rewritten only moved)
    - Deno.json (not yet rewritten only moved)
    - The Standard Library
    - Testing
    - Debugging
    - Workspaces and monorepos
    - Web Frameworks (not yet rewritten only moved)
    - Installation (not yet rewritten only moved)
    - Stability

Remove old files that are incorporated into new structure
Redirect links


runtime/fundamentals/standard_library.md
```diff
@@ -0,0 +1,81 @@
...
+npx jsr add @std/fs @std/path
+```
+
+Will add those packages to your `package.json`:
+
+```json
+{
+  "dependencies": {
+    "@std/fs": "^0.224.0",
+    "@std/path": "^0.224.0"
```
marvinhagemeister, line 76: Output is different + using 1.x versions

```suggestion
    "@std/fs": "npm:@jsr/std__fs@^1.0.2",
    "@std/path": "npm:@jsr/std__path@^1.0.3"
```

runtime/fundamentals/standard_library.md
```diff
@@ -0,0 +1,81 @@
...
+import { join } from "@std/path";
+
+await copy("foo.txt", join("dist", "foo.txt"));
+```
+
+Alternatively, you can import modules directly with the `jsr:` specifier:
+
+```js
+import { copy } from "jsr:@std/fs@^0.224.0";
+import { join } from "jsr:@std/path@^0.224.0";
```
marvinhagemeister, line 54: Just using latest versions because the std recently was marked as stable with 1.x releases

```suggestion
import { copy } from "jsr:@std/fs@^1.0.2";
import { join } from "jsr:@std/path@^1.0.3";
```

runtime/fundamentals/standard_library.md
```diff
@@ -0,0 +1,81 @@
...
+deno add @std/fs @std/path
+```
+
+Your updates your deno.json import map will be updated to include those imports:
+
+```json
+{
+  "imports": {
+    "@std/fs": "jsr:@std/fs@^0.224.0",
+    "@std/path": "jsr:@std/path@^0.224.0"
```
marvinhagemeister, line 36: Just using latest versions because the std recently was marked as stable with 1.x releases

```suggestion
    "@std/fs": "jsr:@std/fs@^1.0.2",
    "@std/path": "jsr:@std/path@^1.0.3"
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+---
+
+Deno provides a built-in test runner for writing and running tests in both
+JavaScript and TypeScript. This makes it easy to ensure your code is reliable
+and functions as expected without needing to install any additional dependencies
+or tools.
+
+The Deno test runner fully supports typescript and async tests and allows for
+fine-grained control over permissions for each test, enhancing your code
+security.
```
marvinhagemeister, line 22: We already mention that the test runner supports TS in the paragraph just before this one. I'm not sure if we explicitly need to mention "async tests" as a feature. I don't know of a test runner which doesn't have this. I like the part about passing individual permissions to tests though, we should keep that. Can't come up with a good phrasing right now.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+
+Deno.test("add function adds two numbers correctly", () => {
+  const result = add(2, 3);
+  expect(result).toBe(5);
+});
+
+Deno.test("add function handles negative numbers", () => {
+  const result = add(-2, -3);
+  expect(result).toBe(-5);
+});
```
marvinhagemeister, line 79: Let's remove one of the test cases. I think one is enough to show how expect can be used.

```suggestion
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+  permissions: { read: true },
+  fn: () => {
+    const data = Deno.readTextFileSync("./somefile.txt");
+    assertEquals(data, "expected content");
+  },
+});
+```
+
+If you prefer a "jest-like" `expect` style, the Deno standard library provides
+an `expect` function that can be used in place of `assertEquals`:
```
marvinhagemeister, line 65: Let's link to the expect docs on JSR.

```suggestion
If you prefer a "jest-like" `expect` style, the Deno standard library provides
an [`expect`](https://jsr.io/@std/expect) function that can be used in place of `assertEquals`:
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+Deno allows you to run specific tests or groups of tests using the `--filter`
+option on the command line. This option accepts either a string or a pattern to
+match test names.
+
+Consider the following tests:
+
+```ts
+Deno.test({ name: "my-test", fn: myTest });
+Deno.test({ name: "test-1", fn: test1 });
+Deno.test({ name: "test-2", fn: test2 });
```
marvinhagemeister, line 136: I feel like we should avoid the object-style arguments as they're only needed when you want to pass permissions.

```suggestion
Deno.test("my-test", () => {});
Deno.test("test-1", () => {});
Deno.test("test-2", () => {});
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+### Filtering by string
+
+To run all tests that contain the word "test" in their names, use:
+
+```sh
+deno test --filter "test" tests/
+```
+
+This command will execute `my-test`, `test-1`, and `test-2` because they all
+include "test" in their names.
```
marvinhagemeister, line 148: I feel like we should change the example because after filtering we still run _all_ tests. We should have an example for filtering where we only run a subset of the tests, not all of them.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+```json
+{
+  "test": {
+    "exclude": ["out/"]
+  }
+}
+```
+
+Then running `deno test` in the same directory tree as the configuration file
+will take these options into account.
```
marvinhagemeister, line 195: The configuration options will _always_ be taken into account, so this sentence is wrong.

```suggestion
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+
+## Test definition filtering
+
+Deno provides two options for filtering tests within the test definitions
+themselves: ignoring tests and focusing on specific tests.
```
marvinhagemeister, line 200: I feel like we should avoid the word `filter` here. We already talked about filtering in the prior section so it's a bit confusing why we're talking about this again.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+
+### Filtering Out (Ignoring Tests)
```
marvinhagemeister, line 202: I think it's much better known as "ignoring or skipping" tests. I wonder if it's better to just say that instead of "Filtering out"

```suggestion
### Ignoring/Skipping Tests
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+
+You can ignore certain tests based on specific conditions using the `ignore`
+boolean in the test definition. If `ignore` is set to `true`, the test will be
+skipped. This is useful, for example, if you only want a test to run on a
```
marvinhagemeister, line 206: We should add a sentence that it's still shown as being skipped in the test runner output. Maybe even include the output of the terminal directly here as an example, or later after the next code snippet.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+```ts
+Deno.test({
+  name: "do macOS feature",
+  ignore: Deno.build.os !== "darwin", // This test will be ignored if not running on macOS
+  fn() {
+    // do MacOS feature here
+  },
+});
+```
+
```
marvinhagemeister, line 218: Add an example for the shorthand syntax, for when you don't want to pass arguments and just skip tests for any reason:

```ts
Deno.test.ignore("my test", () => {
  // some test code
})
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+### Filtering In (Only Running Specific Tests)
+
+If you want to focus on a particular test and ignore the rest, you can use the
+`only` option. This tells the test framework to run only the tests with only set
+to true. Multiple tests can have this option set. However, if any test is
+flagged with only, the overall test run will always fail, as this is intended to
+be a temporary measure for debugging.
+
```
marvinhagemeister, line 226: Add shorthand syntax which will likely be used much more frequently:

```ts
Deno.test.only("my test", () => {
  // some test code
})
```

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+set of functions to help you write tests that involve spying, mocking, and
+stubbing. Check out the [documentation on JSR](https://jsr.io/@std/testing) for
+more information on each of these utilities.
+
+## Assertions
+
+To help developers write tests, the
+[Deno standard library](/runtime/fundamentals/standard_library/) comes with a
+built-in assertions module which can be imported from `jsr:@std/assert`. Check
+out the [documentation on JSR for its usage](https://jsr.io/@std/assert).
```
marvinhagemeister, line 287: We already showed examples with the `assert` and `expect` API earlier in the document. Wondering if this section can be removed.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+
+## Coverage
+
+Deno will collect test coverage into a directory for your code if you specify
+the `--coverage` flag when starting `deno test`. This coverage information is
+acquired directly from the V8 JavaScript engine, ensuring high accuracy.
+
+This can then be further processed from the internal format into well known
+formats with the [`deno coverage`](/runtime/manual/tools/coverage/) tool.
```
marvinhagemeister, line 296: We should add a screenshot of our pretty html output to show how cool this is!

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+With the [@std/testing/bdd](https://jsr.io/@std/testing/doc/bdd/~) module you
+can write your tests in a familiar format for grouping tests and adding
+setup/teardown hooks used by other JavaScript testing frameworks like Jasmine,
+Jest, and Mocha.
+
+The `describe` function creates a block that groups together several related
+tests. The `it` function registers an individual test case. Check out the
+[documentation on JSR](https://jsr.io/@std/testing/doc/bdd/~) for more
+information on these functions and hooks.
+
```
marvinhagemeister, line 309: Add paste-able example code snippet here.

runtime/fundamentals/testing.md
```diff
@@ -0,0 +1,757 @@
...
+snapshots are serialized representations of the original values and are stored
+alongside the test files.
+
+Snapshot testing enables catching a wide array of bugs with very little code. It
+is particularly helpful in situations where it is difficult to precisely express
+what should be asserted, without requiring a prohibitive amount of code, or
+where the assertions a test makes are expected to change often.
+
+---
+
```
marvinhagemeister, line 431: Everything below here is about the LSP protocol integration. I feel like this should be a separate page about "integrating with the LSP" or something. It's not relevant for folks wanting to write tests with Deno. The only nice tip is that you can pass additional arguments to tests started from one of the code lenses by adding it in VSCode's settings file.

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+
+TypeScript is a first class language in Deno, just like JavaScript or
+WebAssembly, so you can run or import TypeScript without installing anything
+more than the Deno CLI.
+
+## The TS Compiler
+
+Deno has a built-in TypeScript compiler, which is used to compile TypeScript
+code to JavaScript. No extra installation or config is required making your
+development process simpler.
```
marvinhagemeister, line 21: I feel like we're repeating ourselves a bit: "...you can run (...) TypeScript..." "Deno has a built-in TypeScript compiler, which is used to compile TypeScript code to JavaScript".

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+TypeScript compiler cache: "/path/to/cache/deno/gen"
+```
+
+If you were to look in the cache, you would see a directory structure that
+mimics your source directory structure and individual `.js` and `.meta` files.
+The `.js` file is the transformed source file while the `.meta` file contains
+meta data we want to cache about the file. The metadata contains a _hash_ of the
+source module that helps Deno to manage cache invalidation. You might also see a
+`.buildinfo` file, a TypeScript compiler incremental build information file,
+used to help speed up type checking.
```
marvinhagemeister, line 47: Is this relevant for Deno users? I feel like we're going deep into implementation details here. For users it doesn't matter if Deno transpiles it internally to JS or if we do it any other way. The main takeaway is that Deno runs TS natively.

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+# or also type check remote modules and npm packages
+deno check --all module.ts
+```
+
+:::note
+
+Type checking can take a significant amount of time, especially if you are
+working on a codebase where you are making a lot of changes. Deno optimizes type
+checking, but it still comes at a cost. Therefore, **by default, TypeScript
+modules are not type-checked before they are executed**.
```
marvinhagemeister, line 78: We should add a note about `deno test`. Because there the type checking is enabled by default and can be disabled via `--no-check`

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+
+:::
+
+When using the `deno run` command, Deno will only transpile the module before
+executing it, ignoring any potential type-related issues. In order to perform a
+type check of the module before execution occurs, you can use the `--check` flag
+with `deno run`:
```
marvinhagemeister, line 85: We're talking about implementation details here. Maybe we can just say something like: "When using the `deno run` command, Deno will skip type-checking the code and run it directly."

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+
+```shell
+deno run --check module.ts
+# or also type check remote modules and npm packages
+deno run --check=all module.ts
+```
+
+Deno treats type checking issues as terminal. If you have used the `--check`
```
marvinhagemeister, line 93: Might be a "me" think. I'm always a bit confused by the `terminal` word given that it has other meanings in programming too. Maybe we should use simpler phrasing here like:

"When Deno encounters a type error, the process with exit before executing the code."

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+| `application/javascript`   | JavaScript (with path extensions influence)                 |
+| `text/javascript`          | JavaScript (with path extensions influence)                 |
+| `application/ecmascript`   | JavaScript (with path extensions influence)                 |
+| `text/ecmascript`          | JavaScript (with path extensions influence)                 |
+| `application/x-javascript` | JavaScript (with path extensions influence)                 |
+| `application/node`         | JavaScript (with path extensions influence)                 |
+| `text/jsx`                 | JSX                                                         |
+| `text/tsx`                 | TSX                                                         |
+| `text/plain`               | Attempt to determine that path extension, otherwise unknown |
+| `application/octet-stream` | Attempt to determine that path extension, otherwise unknown |
```
marvinhagemeister, line 139: Should we mention JSON here? Not sure, given that this is the TS page.

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Mixing JavaScript and TypeScript
+
+By default, Deno does not type check JavaScript. Check out the documentation on
+[configuring your project](./configuration.md) to find out more about how to
+configure your project to do this. Deno does support JavaScript importing
+TypeScript and TypeScript importing JavaScript, in complex scenarios.
```
marvinhagemeister, line 146: Drop the "in complex scenarios". We also support it in non-complex scenarios

runtime/fundamentals/ts_support.md
```diff
@@ -0,0 +1,337 @@
...
+### Internals
+
+- Deno generates a module graph by parsing the root module and its dependencies,
+  filling two potential slots: the code slot and the type slot.
+
+- During type checking, Deno offers the type slot to the TypeScript compiler if
+  it is filled, otherwise, it offers the code slot.
+
+- This ensures that when importing a `.d.ts` module or using alternative type
+  modules for JavaScript code, the correct types are provided to TypeScript.
```
marvinhagemeister, line 337: Not sure if we need this section.

runtime/manual/node/npm_specifiers.md
```diff
@@ -41,7 +41,8 @@ import chalk from "npm:chalk@5";
 ```
 
 Some packages do not though, but you can specify their types with a
-[`@deno-types`](../advanced/typescript/types.md) directive. For example, using a
+[`@deno-types`](./runtime/fundamentals/ts_support.md) directive. For example,
```
marvinhagemeister, line 44: Nit: Wondering if we should rename that file to `typescript.md` instead of `ts_support.md`. Might be better for SEO.

runtime/fundamentals/testing.md
thisisjofrank: I've moved it to its own page for now in tools, which is where the CLI reference is, when we move that all over to the new information architecture we can reconsider where it might be best to go.

runtime/fundamentals/testing.md
thisisjofrank: Good point

go.json
```diff
@@ -1,15 +1,15 @@
...
-  "--allow-ffi": "/runtime/manual/basics/permissions/#ffi-(foreign-function-interface)",
-  "--allow-hrtime": "/runtime/manual/basics/permissions/#high-resolution-time",
-  "--allow-net": "/runtime/manual/basics/permissions/#network-access",
-  "--allow-read": "/runtime/manual/basics/permissions/#file-system-read-access",
-  "--allow-run": "/runtime/manual/basics/permissions/#running-subprocesses",
-  "--allow-sys": "/runtime/manual/basics/permissions/#system-information",
-  "--allow-write": "/runtime/manual/basics/permissions/#file-system-write-access",
-  "--allow-all": "/runtime/manual/basics/permissions/#all-permissions",
-  "--unsafely-ignore-certificate-errors": "/runtime/manual/basics/permissions/#certification-errors"
+  "permissions": "/runtime/fundamentals/permissions/",
```
marvinhagemeister, line 2: This doesn't seem to exist

go.json
```diff
@@ -1,15 +1,15 @@
...
+  "config": "/runtime/fundamentals/configuration/",
+  "ide": "/runtime/getting_started/setup_your_environment/#setting-up-your-editor%2Fide",
+  "--allow-env": "/runtime/fundamentals/permissions/#environment-access",
+  "--allow-ffi": "/runtime/fundamentals/permissions/#ffi-(foreign-function-interface)",
```
marvinhagemeister, line 6: does not exist

marvinhagemeister (approved):

PR #759, opened 2024-08-28
docs: add error messages to the style guide

irbull:
Adds a section to the style guide that explains how to write error messages for the TypeScript / JavaScript parts of the Deno codebase.

This style guide was used in the `std` and seems to be a good fit for the runtime as well. It can be updated as we refactor the existing error messages.

https://github.com/denoland/deno/issues/25269


thisisjofrank (approved):
LGTM

PR #763, opened 2024-08-29
fix: add breadcrumb microdata

marvinhagemeister:


iuioiua (approved):
Can confirm this matches up to the microdata example in https://schema.org/BreadcrumbList. LGTM.

PR #766, opened 2024-08-30
docs: add `--unstable-http`

mattvr:
This updates to include `--unstable-http` on the unstable page, which isn't otherwise explained in the docs. (reference: https://github.com/denoland/deno/blob/main/runtime/js/90_deno_ns.js#L293)


marvinhagemeister (approved):
Sweet, thanks!

PR #778, opened 2024-09-02
feat: close menu with escape key

ryuapp:
On many websites, we can close modals and sidebars by pressing the escape key.
This is a shortcut for user comfort as well as for accessibility.


thisisjofrank (approved):
Nice, thank you

PR #781, opened 2024-09-03
fix: circular redirect

marvinhagemeister:


thisisjofrank:
This page should also be added to the _data.ts file

thisisjofrank (approved):

PR #785, opened 2024-09-04
docs: fix codeblock formatting

marvinhagemeister:
Noticed this small error where too many backticks were used.


marvinhagemeister:
Looks like this was already resolved through another PR

PR #806, opened 2024-09-05
Lsp

thisisjofrank:
moving lsp to references and putting testing in same page


runtime/reference/lsp.md
```diff
@@ -1,14 +1,224 @@
...
+language server:
+
+- `deno.enable`
+- `deno.enablePaths`
+- `deno.codeLens.test`
+
+Deno analyzes these settings at multiple points in the language server process.
+First, when the `initialize` request from the client, the
+`initializationOptions` will be assumed to be an object that represents the
+`deno` namespace of options. For example, the following value will enable Deno
```
marvinhagemeister, line 66: Something is off with that sentence. It feels like a verb or something is missing in the `initialize` part

marvinhagemeister (approved):
LGTM, one sentence feels a bit odd, but that's all. It's much better this way of combining the LSP pages.

runtime/reference/lsp.md
thisisjofrank, line 66:
```suggestion
First, when the `initialize` request arrives from the client, the
`initializationOptions` will be assumed to be an object that represents the
`deno` namespace of options. For example, the following value will enable Deno
```

PR #822, opened 2024-09-10
moving files into reference from /runtime/manual, this needs further …

thisisjofrank:
…cleaning, but this is an initial sweep


runtime/contributing/building_from_source.md
```diff
@@ -11,8 +12,8 @@ chapter).
 
 ## Cloning the Repository
 
-> Deno uses submodules, so you must remember to clone using
-> `--recurse-submodules`.
+:::note Deno uses submodules, so you must remember to clone using
+`--recurse-submodules`. :::
```
marvinhagemeister, line 16: Convention around the container syntax expects this to be defined via newlines. That way you can extend this to pass additional attributes like custom titles which we don't support at the moment (but might in the future).

```md
:::note title="foo bar"

Some content

:::
```

```suggestion
:::note

Deno uses submodules, so you must remember to clone using
`--recurse-submodules`.

:::
```

runtime/contributing/style_guide.md
```diff
@@ -1,13 +1,15 @@
...
+- /runtime/manual/contributing/style_guide/
+- /runtime/manual/references/contributing/style_guide/
 ---
 
-> ⚠️ Note that this is the style guide for **internal runtime code** in the Deno
-> runtime, and in the Deno standard library. This is not meant as a general
-> style guide for users of Deno.
+:::note Note that this is the style guide for **internal runtime code** in the
+Deno runtime, and in the Deno standard library. This is not meant as a general
+style guide for users of Deno. :::
```
marvinhagemeister, line 10:
```suggestion
:::note

Note that this is the style guide for **internal runtime code** in the
Deno runtime, and in the Deno standard library. This is not meant as a general
style guide for users of Deno.

:::
```

runtime/contributing/style_guide.md
```diff
@@ -356,78 +358,78 @@ export function foo(): string {
...
 User-facing error messages raised from JavaScript / TypeScript should be clear,
 concise, and consistent. Error messages should be in sentence case but should
 not end with a period. Error messages should be free of grammatical errors and
 typos and written in American English.
 
-> ⚠️ Note that the error message style guide is a work in progress, and not all
-> the error messages have been updated to conform to the current styles. Error
-> message styles that should be followed:
+:::note Note that the error message style guide is a work in progress, and not
+all the error messages have been updated to conform to the current styles. :::
```
marvinhagemeister, line 369:
```suggestion
:::note

Note that the error message style guide is a work in progress, and not
all the error messages have been updated to conform to the current styles.

:::
```

runtime/_data.ts
```diff
@@ -197,16 +205,16 @@ export const sidebar = [
       {
         label: "Contributing to Deno",
         items: [
-          "/runtime/manual/references/contributing/",
```
marvinhagemeister, line 200: Missing redirect

marvinhagemeister (approved):

PR #826, opened 2024-09-10
moving typescript files and compressing into one doc in reference

thisisjofrank:


runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+
+```json
+{
+  "compilerOptions": {
+    "checkJs": true
+  }
+}
+```
+
+And then setting the `--config` option on the command line.
```
marvinhagemeister, line 41: Nit: This reads a bit odd as a sentence alone. It makes sense if it would be part of the previous one.

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+// @ts-nocheck
+```
+
+## Renaming JS files to TS files
+
+While this might work in some cases, it has some severe limits in Deno. This is
+because Deno, by default, runs type checking in what is called _strict mode_.
+This means a lot of unclear or ambiguous situations where are not caught in
+non-strict mode will result in diagnostics being generated, and JavaScript is
+nothing but unclear and ambiguous when it comes to types.
```
marvinhagemeister, line 77: A bit worried about the `it has some sever limits in Deno` phrasing. Makes it sound like it's something a user shouldn't do. We want to encourage users to go with TS. Maybe something like:

```suggestion
TypeScript files benefit from the TypeScript compiler being able to do more thorough safety checks of your code. This is often referred to as _strict mode_. When you rename a `.js` file to `.ts` you'll might see new type errors that you TypeScript wasn't able to detect before.
```
```

File: runtime/reference/ts_config_migration.md [hunk 3]
```diff
@@ -0,0 +1,501 @@
...
+
+## Configuring TypeScript in Deno
+
+While TypeScript offers many configuration options, which can be daunting if
+you're just starting out with TS. Deno aims to simplify using TypeScript,
+instead of drowning you in countless settings, Deno configures TypeScript to
+**just work** out of the box. No extra configuration headaches required!
```
marvinhagemeister, line 84: The first sentence reads a bit odd:

> While TypeScript offers many configuration options, which can be daunting if
you're just starting out with TS.

Maybe we can start with the good parts:

```suggestion
TypeScript is configured out of the box as you'd expect with Deno. For most projects you'll never have to touch any TypeScript related configuration options. If you do run into the need to change some of the TypeScript's compiler options, you can do so via `deno.json`.
```

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+For a full list of compiler options and how they affect TypeScript, please refer
+to the
+[TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/compiler-options.html).
+
+## Using the "lib" property
+
+Deno includes unique built-in libraries not found in other platforms (like
+`tsc`). This can sometimes pose challenges when writing code meant to run in
+both Deno and a browser. In such cases, you can tweak Deno’s behavior using the
+"lib" property within your `compilerOptions`.
```
marvinhagemeister, line 147: Nit: There shouldn't be "challenges" with Deno when it comes to the wording. Maybe something like:

```suggestion
If you're working on a project that ships code to multiple runtimes, like browsers for example, you can tweak the default types via the "lib" property within the `compilerOptions`.
```

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+  libraries.
+- `"deno.worker"` - This is the library used when checking a Deno web worker
+  script. For more information about web workers, check out
+  [Type Checking Web Workers](./types.md#type-checking-web-workers).
+- `"dom.asynciterable"` - TypeScript currently does not include the DOM async
+  iterables that Deno implements (plus several browsers), so we have implemented
+  it ourselves until it becomes available in TypeScript.
+
+These are common libraries that Deno doesn't use, but are useful when writing
+code that is intended to also work in another runtime:
```
marvinhagemeister, line 169:
```suggestion
These are common libraries that are not enabled by default, but are useful when writing
code that is intended to also work in another runtime:
```

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+```
+
+Typically, when you use the `"lib"` option in TypeScript, you need to include an
+"es" library as well. In the case of `"deno.ns"` and `"deno.unstable"`, they
+automatically include `"esnext"` when you bring them in.
+
+The biggest "warning" when doing something like this, is that the type checking
+is significantly looser, and there is no way to validate that you are doing
+sufficient and effective feature detection in your code, which may lead to what
+could be trivial errors becoming runtime errors.
```
marvinhagemeister, line 227: Not sure what this is referring to. I might be missing context.

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+
+:::note
+
+If you get type errors like `**cannot find`document`or`HTMLElement`**, it is
```
marvinhagemeister, line 231:
```suggestion
If you get type errors like `**cannot find `document` or `HTMLElement`**, it is
```

runtime/reference/ts_config_migration.md
```diff
@@ -0,0 +1,501 @@
...
+}
+```
+
+Then when running deno subcommand, you would need to pass the
+`--config path/to/file` argument, or if you are using an IDE which leverages the
+Deno language server, set the `deno.config` setting.
+
+If you also have non-worker scripts, you will either need to omit the `--config`
+argument, or have one that is configured to meet the needs of your non-worker
+scripts.
```
marvinhagemeister, line 440: This seems to duplicate an earlier section about supporting the browser.

marvinhagemeister (commented):
I like the merging into one document. Left some comments, let me know what you think.

runtime/reference/ts_config_migration.md
thisisjofrank: lets just take it out, i don't think its particularly helpful text anyway

runtime/reference/ts_config_migration.md
thisisjofrank: I wasn't sure whether it was worth reiterating that it works for workers too, think we should take it out?

runtime/fundamentals/typescript.md
```diff
@@ -6,9 +6,6 @@ oldUrl:
   - /runtime/manual/typescript/overview/
   - /runtime/manual/getting_started/typescript/
   - /manual/advanced/typescript
-  - /runtime/manual/advanced/typescript/overview/
```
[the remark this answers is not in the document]
thisisjofrank: This should now be in typescript.md

runtime/fundamentals/typescript.md
```diff
@@ -6,9 +6,6 @@ oldUrl:
...
-  - /runtime/manual/advanced/typescript/types/
-  - /runtime/advanced/typescript/configuration/
```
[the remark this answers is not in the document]
thisisjofrank: These are now in ts_config_migration.md

runtime/manual/advanced/faqs.md
```diff
@@ -1,16 +0,0 @@
----
-title: "Frequently Asked Questions"
-oldUrl:
-- /runtime/manual/node/faqs
```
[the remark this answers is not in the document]
thisisjofrank: This is now in node.md

runtime/manual/advanced/typescript/types.md
```diff
@@ -1,292 +0,0 @@
----
-title: "Types and Type Declarations"
-oldUrl:
- - /runtime/manual/typescript/types/
- - /runtime/fundamentals/types/
```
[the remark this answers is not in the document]
thisisjofrank: These are now in ts_config_migration.md

runtime/reference/ts_config_migration.md
thisisjofrank: Have updated

marvinhagemeister (approved):

PR #833, opened 2024-09-12
updating tutorials and adding new tutorial for init

thisisjofrank:
update hello world, http server and fetch tutorials and add init tutorial

### DIFF
--- a/runtime/_data.ts
+++ b/runtime/_data.ts
@@ -220,22 +220,18 @@ export const sidebar = [
   {
     title: "Tutorials and Examples",
     items: [
-      {
-        label: "Basic Examples",
-        items: [
-          "/runtime/tutorials/hello_world/",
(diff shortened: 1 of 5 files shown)


marvinhagemeister (approved):
This is great! I like the endings with the dino emoji, that's a cute touch!

Only thing I'd change is to get rid of the `## Tutorial` headings. They feel superfluous.

thisisjofrank:
> This is great! I like the endings with the dino emoji, that's a cute touch!
> 
> Only thing I'd change is to get rid of the `## Tutorial` headings. They feel superfluous.

I wanted a way to split the 'intro' content from the 'now do this' content so if you didn't want to read all the explanation you could just jump straight in. But yeah, couldn't come up with a good heading for 'do this bit'.

## Instructions    maybe?

PR #839, opened 2024-09-13
feat: add WebGPU env vars

crowlKats:


thisisjofrank (approved):
Looks like it needs a fmt, but looks good apart from that

PR #841, opened 2024-09-15
KV Key Space: pluralize `users` for consistency

canac:
Pluralize the `users` part of the key like the rest of the page does.


thisisjofrank:
Nice spot, thanks

PR #847, opened 2024-09-16
fix: examples copy button to work clientside

thisisjofrank:
The problem: Lume components don't run in the browser. They are intended to generate static HTML code on building time.
The copy button has an onClick handler that needs to run clientside.

HackyFix: add a click attribute to the button and add the entire codeblock as a data attribute that can be copied on click.

(and also make it so we can have newlines in the description)


examples.page.tsx
```diff
@@ -341,11 +346,12 @@ function SnippetComponent(props: {
...
   return (
     <button
-      onClick={() => navigator?.clipboard?.writeText(props.text)}
       type="button"
       aria-label="Copy code to clipboard"
       title="Copy"
-      class="clean-btn copy-all absolute right-2 top-6 hover:bg-gray-200 text-gray-900 p-2 rounded-md z-10"
+      class="clean-btn copy-all absolute right-2 top-10 hover:bg-gray-200 text-gray-900 p-2 rounded-md z-10"
+      data-content={props.text}
+      onclick="navigator?.clipboard?.writeText(this.dataset['content'])"
```
marvinhagemeister, line 354: This can be removed

marvinhagemeister (approved):

PR #849, opened 2024-09-16
chore: run build on CI

marvinhagemeister:
Run the `build` task also on CI which would've caught the syntax error found in https://github.com/denoland/docs/pull/848 that was introduced by https://github.com/denoland/docs/pull/843

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,3 +12,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: denoland/setup-deno@v1
       - run: deno fmt --check
+      - run: deno task build


lucacasonato:
Can't we just make the `Deploy` check required?

marvinhagemeister:
Wouldn't that cause a deployment?

lucacasonato:
I mean we always run the `Deploy / deploy` step on PRs, no?

lucacasonato:
If you look at the PR, you see we landed that even though CI was red: https://github.com/denoland/docs/pull/843. If we prevent that, then this is not needed.

marvinhagemeister:
I don't have the permissions to change the checks or which ones are required.

lucacasonato:
I made them required now.

PR #855, opened 2024-09-17
Restyle "usage in deno" node reference

donjo:
An idea for a restyled usage block makes it feel more like a labeled content block than a random floating header that's missing paragraph text below

Screenshot 2024-09-17 at 10 45 31 AM


thisisjofrank (approved):
LGTM !important

PR #870, opened 2024-09-20
Update hex-base64-encoding.ts

pdynarowski:
Updated because current doesnt work


kt3k:
Sorry for the delay in review.

This change still seems having an error in import statement.

We already fixed the errors in https://github.com/denoland/docs/pull/1114/files

Thanks for your contributions anyways.

PR #878, opened 2024-09-20
Rename 'Reference' to 'API reference'

bartlomieju:
Main difference is in the top navbar where "Reference" was changed to "API reference".


donjo:
I'm fine with this but happy to be overruled by @thisisjofrank

donjo (approved):

thisisjofrank (approved):
👍

PR #880, opened 2024-09-23
Add append example to writing-files.ts

dandv:
Super common use case that somehow was missing.


thisisjofrank (approved):

thisisjofrank:
Thank you for adding this!
If you'd like some stickers as an additional thank you, drop us an email at docs@deno.com!

PR #889, opened 2024-09-24
updated websockets tutorial and updated bits of other tutorials

thisisjofrank:
waiting on demo being on denoland


runtime/tutorials/chat_app.md
```diff
@@ -4,144 +4,121 @@ oldUrl:
...
 ---
 
-In this tutorial we'll create a simple chat app using Deno. Our chat app will
-allow multiple chat clients connected to the same backend to send group messages
-through web sockets. After a client chooses a username, they can then start
-sending group messages to other online clients. Each client also displays the
-list of currently active users.
+WebSockets are a powerful tool for building real-time applications. They allow
+for bidirectional communication between the client and server without the need
+for constant polling. A frequent use case for web sockets is chat applications.
```
marvinhagemeister, line 9:
```suggestion
for constant polling. A frequent use case for web sockets are chat applications.
```

runtime/tutorials/chat_app.md
```diff
@@ -4,144 +4,121 @@ oldUrl:
...
-      for (const username of data.usernames) {
-        userListHtml += `<div> ${username} </div>`;
-      }
-      document.getElementById("users").innerHTML = userListHtml;
-      break;
+app.use(router.routes());
+app.use(router.allowedMethods());
+app.use(async (context) => {
+  await context.send({
+    root: `${Deno.cwd()}/`,
```
marvinhagemeister, line 54: Should be the same thing.
```suggestion
    root: Deno.cwd(),
```

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
 Now if you visit [http://localhost:8080](http://localhost:8080/) you will be
-able to start a chat session. You can open 2 simultaneous windows and try
-chatting with yourself.
+able to start a chat session. You can open 2 simultaneous tabs and try chatting
+with yourself.
+
+![Chat app UI](./images/websockets.gif)
 
-## Example Code
+TODO: link to repo
```
marvinhagemeister, line 286: TODO

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
+            <h2>Users online</h2>
+            <ul id="users"></ul>
+        </aside>
+        <main>
+            <div id="conversation"></div>
+            <form>
+                <textarea id="data" placeholder="send message"></textarea>
+                <button type="submit" id="send">Send ᯓ✉︎</button>
+            </form>
+        </main>
```
marvinhagemeister, line 180: Let's move the script tag to just before the closing `</body>` tag. That way we don't need to explain the `load` event. We should also use `type="module"` for strict mode and stuff.

```suggestion
    </head>
    <body>
        <header>
            <h1>🦕 Deno Chat App</h1>
        </header>
        <aside>
            <h2>Users online</h2>
            <ul id="users"></ul>
        </aside>
        <main>
            <div id="conversation"></div>
            <form>
                <textarea id="data" placeholder="send message"></textarea>
                <button type="submit" id="send">Send ᯓ✉︎</button>
            </form>
        </main>
        <script src="/public/app.js" type="module"></script>
```

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
+    </body>
+</html>
+```
+
+### CSS
+
+If you'd like to style your chat app, create a `style.css` file in the `public`
+folder and add the following code:
+
+TODO: link to css
```
marvinhagemeister, line 190: TODO

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
-  const socket = await ctx.upgrade();
-  const username = ctx.request.url.searchParams.get("username");
-  if (connectedClients.has(username)) {
-    socket.close(1008, `Username ${username} is already taken`);
-    return;
+We'll set up the client side JavaScript in an `app.js` file, you'll have seen it
+linked in the HTML we just wrote. In the `public` folder and add an `app.js`
+file with the following code:
+
+```js
```
marvinhagemeister, line 198:
````suggestion
```js title="app.js"
````

File: runtime/tutorials/chat_app.md [hunk 7]
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
+    listItem.textContent = username;
+    userList.appendChild(listItem);
+  }
+}
+
+// Add a new message to the conversation
+function addMessage(username, message) {
+  const conversation = document.getElementById("conversation");
+  const messageDiv = document.createElement("div");
+  messageDiv.innerHTML = `<span>${username}</span> ${message}`;
```
marvinhagemeister, line 236: I know that this is a beginner tutorial, but we should never put user input directly into HTML strings. Doing so makes the visitor to that website vulnerable to injection attacks. Both the `username` and `message` parameter can be abused here by passing malicious HTML strings.

```suggestion
  const userLabel = document.createElement("span");
  userLabel.textContent = username;
  messageDiv.appendChild(userLabel);
  messageDiv.appendChild(` ${message}`)
```

marvinhagemeister: Looking at all the manual `document.createElement` calls I wonder if it would be better to add a `<template>` tag in the HTML and clone that instead. Would be likely less JS code.

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
+  conversation.prepend(messageDiv);
+}
+
+// Set up event listener for sending messages
+window.onload = () => {
```
marvinhagemeister, line 241: If the `<script>` tag is moved to before the closing `</body>` then this won't be needed anymore.

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
+  const inputElement = document.getElementById("data");
+  const sendButton = document.getElementById("send");
+
+  // Listen for Enter key press
+  inputElement.addEventListener("keypress", (e) => {
```
marvinhagemeister, line 246: The `keypress` event has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) for a while.

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
 
-console.log("Listening at http://localhost:" + port);
-await app.listen({ port });
+  // Listen for button click
+  sendButton.addEventListener("click", (e) => {
+    e.preventDefault();
+    const message = inputElement.value;
+    inputElement.value = "";
+    socket.send(JSON.stringify({ event: "send-message", message }));
+  });
```
marvinhagemeister, line 260: We should put the input element in a form and do normal form submission. That way we don't need to duplicate this logic and only need one `submit` listener.

runtime/tutorials/chat_app.md
```diff
@@ -163,73 +140,154 @@ function broadcast_usernames() {
...
 
-```sh
-deno run --allow-read --allow-net server.js
+## Running the server
+
+To run the server we'll need to grant the necessary permissions to Deno. In your
+`deno.json` file, update the `dev` task to allow read and network access:
+
+```sh title="deno.json"
+"dev": "deno run --allow-net --allow-read --watch main.ts"
```
marvinhagemeister, line 277: Maybe we should do a diff here to see what's changed compared to the old command.

runtime/tutorials/chat_app.md
thisisjofrank: Updated

runtime/tutorials/chat_app.md
thisisjofrank: I've put a defer on it

runtime/tutorials/chat_app.md
thisisjofrank: Yeah, that's a much nicer idea

runtime/tutorials/chat_app.md
```diff
@@ -1,235 +1,303 @@
...
-        <div id="users"></div>
-        <hr class="visible-xs visible-sm" />
-      </div>
-      <div>
-        <input id="data" placeholder="send message" />
-        <hr />
+<head>
+    <title>Deno Chat App</title>
+    <link rel="stylesheet" href="/public/style.css">
+    <script defer src="/public/app.js"></script>
```
marvinhagemeister, line 179:
```suggestion
    <script defer type="module" src="/public/app.js"></script>
```

runtime/tutorials/chat_app.md
```diff
@@ -1,235 +1,303 @@
...
+
+We'll build a simple UI that shows a text input and a send button and displays
+the sent messages, alongside a list of users in the chat.
+
+### HTML
+
+In your new project directory, create a `public` folder and add an `index.html`
+file and add the following code:
+
+```html title="index.html"
```
marvinhagemeister, line 174: Let's add the doctype to be fully valid html code.

````suggestion
```html title="index.html"
<!DOCTYPE html>
````

marvinhagemeister (approved):
This revision is great! The JS code is much easier to read and there is a lot less needed in this iteration. Good work! Left some minor nit comments

PR #898, opened 2024-09-26
Add JSON body example to http-requests.ts

dandv:
Please squash.

### DIFF
--- a/examples/http-requests.ts
+++ b/examples/http-requests.ts
@@ -40,7 +40,9 @@ resp = await fetch("https://example.com", {
     "Content-Type": "application/json",
     "X-API-Key": "foobar",
   },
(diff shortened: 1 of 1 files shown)


examples/http-requests.ts
```diff
@@ -40,7 +40,9 @@ resp = await fetch("https://example.com", {
     "Content-Type": "application/json",
     "X-API-Key": "foobar",
   },
-  body,
+  body: JSON.stringify({
+    param: "value",
+  }),
```
marvinhagemeister, line 45: Let's drop the `body` variable in line 36 now that it is inlined here.

marvinhagemeister (approved):
Sweet, thanks!

PR #900, opened 2024-09-26
docs: add prefix to `deno add` arguments

marvinhagemeister:
Deno 2 will require prefixes to be present for the package arguments to `deno add` and `deno install`


thisisjofrank (approved):
jsr:@lgtm/gyat

PR #904, opened 2024-09-26
feat: update deno doc and types/node

crowlKats:
Adds:
- Class subitems in overviews
- node typings jsdocs use ESM instead of require


reference_gen/node-docs.ts
```diff
@@ -84,13 +84,18 @@ for (const file of files) {
...
       if (
+        !exportable.wasForgotten() &&
         exportable.getKind() !== ts.SyntaxKind.ImportEqualsDeclaration &&
         !(exportable.getKind() === ts.SyntaxKind.ModuleDeclaration &&
           exportable.getName() === "global")
       ) {
-        exportable.setIsExported(
-          !EXCLUDE_MAP[fileName]?.includes(exportable.getName?.()),
-        );
+        if (exportable.getName?.() === "BuiltInModule") {
```
crowlKats, line 92: this is an annoying workaround, but better for now to exclude it and re-introduce once problems related to it are fixed. causes several issues and have to increase max-heap to 6GB+

marvinhagemeister (approved):

PR #913, opened 2024-09-27
fixed incorrect href for Websocket under Deno API

trace2798:
Currently from the official [docs](https://docs.deno.com/) website, when someone clicks on WebSocket under Deno API, it redirects to a Not-found page. 

Expected Behavior: Should redirect to the correct page.

Reason: a small typo in the href associated with [it](https://github.com/denoland/docs/blob/33fc25bccb53a9bed708966ccd01880e2f3c862f/index.page.tsx#L273). 
Current href: "/api/deno/websocket"
Correct href:  "/api/deno/web-sockets"

image

### DIFF
--- a/index.page.tsx
+++ b/index.page.tsx
@@ -270,7 +270,7 @@ export default function () {
                     },
                     {
                       text: "Websocket",
-                      href: "/api/deno/websocket",
+                      href: "/api/deno/web-sockets",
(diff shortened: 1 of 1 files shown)


thisisjofrank:
Fixed in https://github.com/denoland/docs/pull/915

PR #914, opened 2024-09-27
changes cookies to canvas from the list of Web APIs

trace2798:
Changes the option from Cookies to Canvas. Cookies does not exist in the list of web api in this [page](https://docs.deno.com/api/web/)

When someone clicks on Cookies from the list under [Web APIs](https://docs.deno.com/) , it redirects them to a not-found page. I believe it will be a better option to change it to a link that exists. 

image

### DIFF
--- a/index.page.tsx
+++ b/index.page.tsx
@@ -270,7 +270,7 @@ export default function () {
                     },
                     {
                       text: "Websocket",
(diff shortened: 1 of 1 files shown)


thisisjofrank:
Fixed in https://github.com/denoland/docs/pull/915

PR #931, opened 2024-10-02
feat: generate node compat list from files

crowlKats:


runtime/_data.ts
```diff
@@ -240,3 +242,91 @@ export const sidebar = [
...
+      throw `Invalid status provided in '${dirEntry.name}': ${parsed.status}`;
+    }
+
+    descriptions[dirEntry.name.slice(0, -5)] = parsed;
+  }
+
+  return descriptions;
+}
+
+export async function generateNodeCompatability() {
```
ry, line 308: Maybe add some comments here explaining how this works for future readers

ry (approved):
LGTM

PR #938, opened 2024-10-03
lsp internals page, old url is same as current url -> 301 redirect loop

roelandxyz:
When visiting https://docs.deno.com/runtime/reference/cli/lsp_integration/ there is an eternal redirect loop. 

In that markdown page is an oldUrl specified that is the same as the correct url. This PR just removes that line and the page works again. (after the 301 cache is cleared in the browser)


bartlomieju (approved):
Thanks!

PR #951, opened 2024-10-05
Update default values of `noImplicitOverride` and `useUnknownInCatchVariables`

kt3k:
These values have been changed in https://github.com/denoland/deno/pull/25695 and https://github.com/denoland/deno/pull/25465


thisisjofrank (approved):
👍

PR #958, opened 2024-10-07
fix: make node compat notices more accurate

crowlKats:


reference_gen/node_descriptions/zlib.yaml
```diff
@@ -1 +1,7 @@
-status: good
+status: partial
+symbols:
+  Options: This class is not supported.
+  BrotliOptions: This class is not supported.
+  BrotliCOmpress: This class is not supported.
```
bartlomieju, line 5:
```suggestion
  BrotliCompress: This class is not supported.
```

bartlomieju (approved):

PR #994, opened 2024-10-12
Update instruction to generate reference docs

kidonng:
There is a new deno task `reference` to generate reference docs since #812.


kt3k (approved):
LGTM. Thanks!

PR #1000, opened 2024-10-13
feat: adds JWT example

shajanjp:
Adds a new example demonstrating how to create and verify JSON Web Tokens (JWT) using the jose library in Deno. The example includes two main functions: createJWT for generating JWTs and verifyJWT for verifying them.

### DIFF
--- a/examples/creating-and-verifying-jwt.ts
+++ b/examples/creating-and-verifying-jwt.ts
@@ -1,0 +1,58 @@
+/**
+ * @title Creating and verifying JWT
+ * @difficulty intermediate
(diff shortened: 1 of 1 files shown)


fwqaaq:
I don't think this is necessary - this should be left to the library's author to handle.

kt3k (approved):
We discussed about including the examples that use 3rd party libraries and we think it's fine as long as the 3rd party lib is established one in the category.

`/x/jose` seems well used. Also it supports Deno well (Their CI runs test cases in Deno).

I think it's reasonable to include this example in our docs.

PR #1005, opened 2024-10-14
update the command for .ts file

Saif-Shines:
Code about is .js file but the command is run for `.ts` file.  Fix: change the filename form `stream.ts` to `stream.js`


marvinhagemeister (approved):
Good spot! Thanks for making a PR 👍

PR #1012, opened 2024-10-15
fix(code-copy-plugin): Remove leading "$" or ">"  from commands when copied

rahulbhooteshwar:
fix(code-copy-plugin): Remove leading "$" or ">"  from commands when copied

- In many command examples, leading  ">" or "$ "signs are used (although the usage is not consistent)
- When copy button is used those leading symbols getting copied
- When copied commands are used by the user, it gives error (it can be frustrating as well).


thisisjofrank (approved):
Oh this is such a good spot. Thank you!

PR #1013, opened 2024-10-15
Add sqlite example

Lioness100:
I thought an example with SQLite might be useful. Please let me know if there's anything I should change!


thisisjofrank (approved):
This is great, thank you! 
Drop us an email at docs@deno.com for some thank you stickers!

PR #1015, opened 2024-10-15
fix: web worker permissions caution block

lucasfernog:
`:::warning` does not render properly, must use `:::caution`


marvinhagemeister (approved):
Good spot! Thanks for making a PR

PR #1016, opened 2024-10-15
Add clarity to documentation for `deno install` and `deno uninstall` CLI subcommands

matthewjnield:
# Commit 1
## Add clarity to documentation for `deno install` CLI subcommand
This commit adds clarity to the documentation of the `deno install` CLI subcommand.

# Commit 2
## Add clarity to documentation for `deno uninstall` CLI subcommand
This commit adds clarity to the documentation of the `deno uninstall` CLI subcommand.

### DIFF
--- a/runtime/reference/cli/install.md
+++ b/runtime/reference/cli/install.md
@@ -13,7 +13,8 @@ command: install
 
 ### deno install
 
-Use this command to install all dependencies defined in `deno.json` and/or
+Use this command with no particular pacakges specified to install all dependencies
+defined in the `imports` field of `deno.json` and/or the `dependencies` field of
 `package.json`.
 
 The dependencies will be installed in the global cache, but if your project has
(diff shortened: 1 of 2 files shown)


kt3k:
These two commands should be identical. If there's any difference, that should be a bug of deno itself. Please file an issue in https://github.com/denoland/deno instead.

matthewjnield:
@kt3k `deno install` and `deno add` are not identical, and one is not an alias for the other; they have different behavior.

`deno add` never does anything more than adding a dependency to the `package.json` configuration file, i.e. it will never actually install the dependency on the machine, but rather simply add the required line of text to `package.json`.

`deno install` actually installs dependencies (those specified to the command, or those in `package.json`) on the machine.

Likewise, `deno uninstall` and `deno remove` are not identical, and one is not an alias for the other; they have different behavior.

`deno remove` never does anything more than removing a dependency from the `package.json` configuration file, i.e. it will never actually uninstall the dependency from the machine, but rather simply remove the required line of text from `package.json`.

`deno uninstall` actually uninstalls dependencies (those specified to the command) from the machine.

The above are the distinctions that I am trying to make clear with the proposed changes to the documentation.

kt3k:
@bartlomieju Is there the difference mentioned above?

PR #1018, opened 2024-10-15
Add dark/light mode

### LINKED ISSUE #817: Docs Dark Mode
Please consider adding a dark mode option to the docs.

Normally I would just use the web extension Dark Reader. However when I use it on your docs the code boxes are unreadable (background becomes white with text mostly illegible).

Thanks for all you do.

davis9001:
Added a dark and light mode, avoided FOUC, used Tailwind and CSS vars().

This might not be the cleanest implementation but as far as I can tell this works and will result in cleaner styling going forward. My biggest criticism of this would be the use of `invert` on the code blocks and the orama search. I tried to avoid that but I couldn't see any way around flashing some styles in some edge cases when trying to use the markdown and orama theme options.

Closes #817


marvinhagemeister:
Thanks for getting the ball rolling on this! Just checked out the branch locally and it looks great!

The only thing

donjo:
@davis9001 Thanks for all of this! Just noting that we're going to chat about this at our docs design meeting later this week and figure out how we want to move forward. Some of the things we'll want to figure out are whether additional colors should be added to our palette to help with the dark mode styles and how we want to handle the use of color inversion (like should we wait to release until we come up with manually selected dark mode colors).

davis9001:
> The only thing I'd change is to keep the blue links. We don't have yellow anywhere in our color scheme. Same for the code blocks.

Excellent point! I added to the pull request to fix the links in the markdown blocks. I haven't fixed the code blocks yet, I think the best long term solution for that would be to override the styling coming from gfm.css.

> Some of the things we'll want to figure out are whether additional colors should be added to our palette to help with the dark mode styles and how we want to handle the use of color inversion (like should we wait to release until we come up with manually selected dark mode colors).

Makes sense! In any case I would definitely recommend having a well documented style guide that includes both light and dark theme color schemes. Do let me know if there is anything I can do to help!

davis9001:
Closing now that #1038 was merged 🥳

PR #1020, opened 2024-10-16
Replace deprecated `tsserver` with `ts_ls` in Neovim 0.6+ lspconfig

gaveen:


gaveen:
Forgot to update the paragraph above the example. Sorry, will update.

PR #1029, opened 2024-10-17
fix: order of install/run

TobyEalden:
The order of the install and run commands is reversed.


marvinhagemeister (approved):
Whoops, good catch. Thanks for making a PR 👍

PR #1036, opened 2024-10-18
Update environment-variables.md

fightbulc:
Added a description on how to update an existing env variable value.

The answer came from orriols on discord:

Today at 2:50 PM
Assuming you are updating the envs in the frontend, you can add a "new" env with the same name as the one you are replacing, and it will replace the env


arnauorriols (approved):
LGTM! Thanks!

fightbulc:
Happy to contribute something back

PR #1038, opened 2024-10-18
Add dark mode

josh-collinsworth:
Takes the [original PR here](https://github.com/denoland/docs/pull/1018) and adds some improvements:

- Removes reliance on inverting colors; sets proper color variables as needed, including:
  - Seting Orama variables to match light or dark mode for the search box and search bar
  - Showing the proper version of the logo for each color theme, instead of inverting the light version in dark mode
- Adjusts the slide-out menu to clean up the jank when switching between light and dark mode (and improve perf)


davis9001:
Ayy nice!

donjo:
This is looking quite good. This was the only piece I could find that needs some adjustment in my standard clicking around.

Screenshot 2024-10-21 at 11 59 36 AM

Also just to throw this on your radar. There is a PR to add the new Orama search box #753 that is now ready for review and they update a few things with how they do their styling. I'll let you figure out how to sequence them but wanted to make sure you know. @crowlKats has more context about the searchbox integration than I do

davis9001:
Something to consider: I'm not sure if y'all want to stick with the way I implemented the mode switcher button, it seems like the standard pattern is to show the moon icon while in dark mode and the sun icon while in light mode... My gut reaction without first looking up the pattern was and is to have the button reflect what the action will do (so the opposite of the current standard pattern).

Just wanted to make sure this is part of the discussion.

davis9001:
Two screenshots of a current little bug in the dark mode on the latest test deployment here:

image
image

Getting the above on a load of the latest deployment here: https://deno-docs-fjz8182244vh.deno.dev/examples/uuids/

Steps to reproduce:
 - Have dark mode enabled in OS (possibly in browser?)
 - Open above deployment build URL for the first time (I used Incognito tab in Chrome to reproduce)
 - After clicking theme toggle button the problem does not persist on page navigation or on new reloads
 - After clicking theme toggle the problem does appear only as a flash on reload using CTRL+SHIFT+R

Just an FYI in case it wasn't already noticed 😉 

I only found it because I was looking something up and wanted to use dark mode 😂

donjo:
@josh-collinsworth ^ any idea about this problem above? I can reproduce it as well but don't know a lot about how the syntax highlighting works

josh-collinsworth:
@donjo @davis9001 That bug should be fixed now. Sorry; I wasn't properly accounting for a situation with an empty `localStorage` value.

While I was at it: I realized `theme` is a very generic key other sites are likely to use, so I changed it to `denoDocsTheme` to avoid any potential collisions.

Looks like we still need to account for the new Orama box, but I don't mind mering this first. (I think we maybe should not do so on a Friday afternoon, however.)

davis9001:
> While I was at it: I realized `theme` is a very generic key other sites are likely to use, so I changed it to `denoDocsTheme` to avoid any potential collisions.

A better key is a good idea but localStorage is namespaced to the current origin, even subdomains on the same domain name should not have access IIRC. 👍

josh-collinsworth:
> A better key is a good idea but localStorage is namespaced to the current origin, even subdomains on the same domain name should not have access IIRC. 👍

You know what – you're right. I was noticing I could access other keys locally, and so I assumed cross-contamination could occur. B I didn't realize until you mentioned this that all the other keys I could access were _also_ from `localhost`. 🤦 

So, totally unnecessary, but still probably not a bad thing to have a more explicit name. Oh well. 😅

pierreoa:
Was checking the preview and it's awesome. Only noticed that the Orama search plugin was still in light mode. Not entirely familiar with the plugin but fixing it [here](https://github.com/denoland/docs/blob/7efc7b509fd90691c9e98a6ab91cad7dce02d55f/search.client.ts#L11) locally resolved it for me.

uncenter:
I don't want to be super nitpicky but I was just taking a look through and noticed some minor issues with the table of contents:

CleanShot 2024-10-27 at 09 17 46

The border/line below the heading text is too bold and the contrast of hovered links is poor.

donjo:
@uncenter I appreciate it! I'm still digging around to find some stray things that need to be cleaned up so I'm hoping to wrap up most of it today. I'm sure there will still be some left when we're done but as long as the majority are taken care of before most people see it I'm fine with that.

davis9001:
I was about to leave the following comment but looking at the code of the last commit I'm guessing that was addressing this 😂 

> Sharing a small glitch I just found (directly after toggling to dark mode from light mode, refreshing fixes the colors it only happens directly going from light to dark).
> 
> Hovering the orama search input in the top nav bar changes it to white:
> image
> 
> In the orama modal the colors are off:
> image
> 
> Steps to reproduce:
>  - After page load: Toggle into dark mode (if you are already in dark mode, toggle to light mode then dark without reloading the page).
>  - Hover the orama search input in the top nav bar (hover background color is white)
>  - Open the orama search (text and some other colors are off)
> 
> I could look into a fix if anyone is stuck on this but I'm guessing you're all on it! 👍

davis9001:
> @davis9001 yeah I am still kind of stuck on that. it seems to reset to the proper colors once you refresh but for the colors you see immediately after toggling your mode, only some of them switch over. I think we're going to look to merge this PR since it hits 99% of the issues and is getting a little unwieldy so we can address any smaller bugs in an independent PR

@donjo I'll at least take a quick 10 minute look at it right now.

davis9001:
@donjo I believe I found a working solution for the orama issue.

In [`search.client.ts`](https://github.com/davis9001/deno-docs/blob/e08f5cca6491387e55b0e82518e07ced74b9dbf6/search.client.ts#L37)

You may want to change which colors are used but using the css vars from the site should be a nice clean way to make this work consistently. It _may_ be better to just override all of these variables in the main site css file but this at least seems to be a working solution for me....

donjo:
@davis9001 I figured out a fix. Basically anywhere there was a color override defined for light, an override had to be defined for dark as well or it would have a hard time falling back to the Orama default.

davis9001:
> @davis9001 I figured out a fix. Basically anywhere there was a color override defined for light, an override had to be defined for dark as well or it would have a hard time falling back to the Orama default.

@donjo Ah yeah... I still think it would be smart to use our variables like I did [here](https://github.com/davis9001/deno-docs/blob/e08f5cca6491387e55b0e82518e07ced74b9dbf6/search.client.ts#L37).

e.g.
```typescript
// ...
    oramaSearchBox.themeConfig = {
      colors: {
        light: {
          "--text-color-primary": "hsla(var(--foreground-primary))",
          "--text-color-secondary": "hsla(var(--foreground-primary))",
          "--text-color-accent": "hsla(var(--foreground-primary))",
// ...
```

donjo (approved):

davis9001:
🥳 @donjo @josh-collinsworth 

Don't forget to close #817

PR #1052, opened 2024-10-21
Closed open quotes on fmt-component

luisfelipeluis49:
[\`deno fmt\` code formating](https://docs.deno.com/runtime/reference/cli/formatter/#supported-file-types)

Notes on Astro, Svelte and Vue file types are `"unstable": ["fmt-component]`, where it should be: `"unstable": ["fmt-component"]` instead, with the closing quotes.

Fixed this on formatter.md, and tested runing `deno task build` and `deno task prod`.


thisisjofrank (approved):
Good spot

PR #1059, opened 2024-10-22
fix: script filename missing `.ts`

staciax:


marvinhagemeister (approved):
Good catch! Thanks for making a PR 👍

PR #1061, opened 2024-10-22
moving file based routing from examples to tutorials 

thisisjofrank:
Needing two files to run this example doesn't work so well in examples, so moving this over to tutorials


marvinhagemeister (approved):
👍

PR #1063, opened 2024-10-23
adding redirects to pages that are 404ing

thisisjofrank:


runtime/reference/migration_guide.md
```diff
@@ -255,7 +255,7 @@ See [deno#9795][deno#9795] for details.
 
 Use the `.close()` method on the resource instead.
 
-<a id="#rid">test</a>
+<a href="#rid" id="#rid">test</a>
```
marvinhagemeister, line 258: Looking at it again:

- The `id` attribute should not start with `#` as that is implied.
- Is it correct if we have a link that references itself?

thisisjofrank: no, that should have been a # on the href and not on the id, clearly I am losing my mind.

PR #1068, opened 2024-10-24
docs: update neon postgres guide to use neon serverless driver on jsr

dhanushreddy291:
Update Neon Postgres guide to use neon serverless driver on jsr


magurotuna (approved):
LGTM

PR #1069, opened 2024-10-24
update nav design flourishes

donjo:
A few design updates for the side and top nav:

- less vertical space on nav link items
- light border on hover to help identify actively hovered items vs current page
- bolded text for current page to provide a secondary means of identifying current page
- adjustments to search box to make it match some of the other grays used on the page
- lightened the divider between top nav item types so it's more clear that its a divider and not an un-rendered link
- reducing icon size for external link so it doesn't stick out so much

Screenshot 2024-10-24 at 12 06 00 PM

### DIFF
--- a/_components/Header.tsx
+++ b/_components/Header.tsx
@@ -60,7 +60,7 @@ export default function Header({
             name="API reference"
             hideOnMobile
           />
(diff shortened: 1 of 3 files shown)


thisisjofrank (approved):
Love this

PR #1073, opened 2024-10-24
Update streaming-files.ts with a simplified loop

fwqaaq:
Replace `do...while` with `while(true)`


ry (approved):
thanks - that's much simpler

PR #1074, opened 2024-10-25
moving components into separate files to make doc less confusing

thisisjofrank:
and deleted old link checker which is now unused


marvinhagemeister (approved):
LGTM, once CI is green.

PR #1077, opened 2024-10-25
Update Lume + Tailwind changes

oscarotero:
Hi, this PR contains two changes:
- Upgrade Lume to the latest version (you were using a develop version from jsdelivr)
- Changed Tailwind implementation: You had a custom plugin to use postcss with tailwind + autoprefixer plugins. Lume has already a postcss plugin with autoprefixer installed by default, so you don't need to create a custom plugin to use tailwind, just use the Lume's postcss plugin configured to use tailwind as an extra plugin.


marvinhagemeister (approved):
Sweet, thanks!

marvinhagemeister:
Looks like I don't have permissions to resolve the merge conflict. Once that's done it can land 👍

Issue #1009, opened 2024-10-15
Add hint about when running `deno install` is needed


marvinhagemeister:
We should add a hint about when running `deno install` is needed on https://docs.deno.com/runtime/fundamentals/configuration/#dependencies

matthewcarlino:
@marvinhagemeister Perhaps a simple "This will require you to run `deno install`." right after the `script.ts` block would suffice?  

I can work on this.

PR #1078, opened 2024-10-26
hint at running  deno install when using a package.json

matthewcarlino:
Added a line of test stating, "Note that this will require you to run `deno install` at `/runtime/fundamentals/configuration/#dependencies`

Resolves [#1009]


thisisjofrank (approved):
Nice addition, thank you!

PR #1079, opened 2024-10-26
Update docs to remove unused context

emctl:
Trying to follow along and wasn't sure where context is even used when following the tutorial from scratch.


thisisjofrank (approved):
Good spot, thank you!

PR #1091, opened 2024-10-29
fix: missing syntax highlighting for jsonc

marvinhagemeister:
Noticed that we weren't highlighting `.jsonc` code snippets on https://docs.deno.com/runtime/fundamentals/configuration/#custom-path-mappings .

This PR fixes that.


thisisjofrank (approved):
👍

PR #1092, opened 2024-10-29
fix(docs): Remove duplicated links in node_api docs

c43721:
Removes the duplicated `fetch` and `File` supported row from the Globals field.

image


donjo (approved):

donjo:
thanks!

PR #1118, opened 2024-11-08
Update ts_config_migration.md

### LINKED ISSUE #1109: Docs include not allowed target property
The documentation at
https://github.com/denoland/docs/blob/main/runtime/reference/ts_config_migration.md#targeting-deno-and-the-browser

says
```
{
  "compilerOptions": {
    "target": "esnext",
    "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]
  }
}
```

However, upon typing that into my `deno.json`, I get
> Property target is not allowed.

image

Does this documentation need to be updated, or is the tooling wrong?

thisisjofrank:
Remove target from example compiler options, fixes #1109

### DIFF
--- a/runtime/reference/ts_config_migration.md
+++ b/runtime/reference/ts_config_migration.md
@@ -193,7 +193,6 @@ using any APIs exclusive to one or the other. In such cases, a typical
 ```json title="deno.json"
 {
   "compilerOptions": {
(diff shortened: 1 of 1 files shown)


philhawksworth (approved):
Jolly good!

PR #1120, opened 2024-11-08
Simplify and stylize feedback module

josh-collinsworth:
| Current | Proposed |
|-|-|
| CleanShot 2024-11-13 at 16 47 59@2x | CleanShot 2024-11-13 at 16 48 42@2x |

| Current | Proposed |
|-|-|
| CleanShot 2024-11-13 at 16 44 50@2x | CleanShot 2024-11-13 at 16 45 20@2x |


lucacasonato:
I think the comment on the email address noting that you can put it in if you want us to contact you seems useful

lucacasonato:
I also like the current "Help us make these docs great!" title - it may entice users to provide feedback

philhawksworth (approved):
This looks like a great design iteration. While there is still some discussion about exact language to be used, I suggest we don't let that block the progress made in having this merged.

We can continue to tweak language

PR #1121, opened 2024-11-09
docs: docs for coverage ignore comments

bcheidemann:
See https://github.com/denoland/deno/pull/26590 and https://github.com/denoland/deno/issues/16626


bcheidemann:
@thisisjofrank Very welcome 😊 I probably should have mentioned this before, but I will probably need to raise a PR to adjust the docs based on the review comments on https://github.com/denoland/deno/pull/26590. I haven't made the docs changes yet as I'm waiting for a re-review and want to get confirmation that the current behaviour is desirable before I update the docs accordingly.

PR #1123, opened 2024-11-09
Update Next.js app setup instructions with Deno

JSStenta:
- Added a step to install dependencies using `deno install` before starting the Next.js app with `deno task dev`.
- Improves clarity and ensures all necessary steps are included in the setup process.


kt3k (approved):
LGTM. Thanks

PR #1128, opened 2024-11-11
Cryptography example demonstrating generation and verification of HMAC

zeecorleone:
Simple example showing how to generate and verify HMAC, which is very common way of ensuring that no data has been tempered when two parties transact/exchange data.


kt3k (approved):
LGTM. Thanks!

PR #1132, opened 2024-11-11
Updated react.md to fix a file path error in example code

aaronjlafitte:
During the example step of including `routeStaticFilesFrom`, in the `main.ts` file, it expects a path of "./util/routeStaticFilesFrom.ts".

But the step fails to mention the creation of that folder. Just made a small modification to the documentation to add the `routeStaticFilesFrom.ts` into a util folder.

Verified that I was able to now successfully run the `deno task serve` command after making this change.


thisisjofrank:
Ah, this has been fixed in another PR by swapping the import from /util/ to /api/. 
Thank you for spotting it though!

PR #1146, opened 2024-11-15
refactor(reference): generate HTML files with the deno/doc jsr package instead of CLI

crowlKats:


kt3k (approved):
Confirmed the same number of files are generated without warning in Deno 2. LGTM

PR #1148, opened 2024-11-16
fix typo in group name for gzip example

dChancellor:
I'm so sorry, at some point during development I deleted an e in the gzip example. This caused a typo in the group name which led to a new group being created.

This PR fixes that typo so it is appropriately grouped.

image


kt3k (approved):
LGTM!

PR #1125, opened 2024-11-10
Update react.md

vijaytv:
the path to routeStaticFilesFrom.ts is inconsistent

PR #1152, opened 2024-11-17
harmonize folder structure in text and code-snippet

SirunGameDev:
so https://github.com/denoland/tutorial-with-react/tree/main/api/util  and code snippet actually works with instruction given in text.
1. file is in repo in util - folder
2. file is not in /api/api/ folder, like code snippets states.
3. text says to create file in api-folder.

https://github.com/denoland/docs/pull/1125 can be closed then. or merge both pr together, so that a good end result is reached.


kt3k (approved):
LGTM

PR #1153, opened 2024-11-19
Update examples/ulid.ts

kt3k:
ULID example doesn't seem working now because of some errors in 'imports' https://docs.deno.com/examples/ulid/

This PR fixes it.


satyarohith (approved):
LGTM

PR #1162, opened 2024-11-19
feat: redirect to previous path if its a 404

crowlKats:
This is an MVP, not the perfect result, but much better than what we currently have (which is nothing)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Deno
         uses: denoland/setup-deno@v2
         with:
-          deno-version: canary
+          deno-version: 2.0.6
           deno-binary-name: "deno_doc"
 
       - name: "Reference: install"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Deno
         uses: denoland/setup-deno@v2
         with:
-          deno-version: 2.0.6
+          deno-version: 2.0.5
           deno-binary-name: "deno_doc"
 
       - name: "Reference: install"
--- a/404/index.page.tsx
+++ b/404/index.page.tsx
@@ -1,0 +1,50 @@
+export const layout = "raw.tsx";
+export const url = "/404";
+import { sidebar } from "../runtime/_data.ts";
+
+const styles = /*css*/ `
+#content nav {
+  padding-inline: 0;
+}
+#content nav > ul {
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Deno
         uses: denoland/setup-deno@v2
         with:
-          deno-version: canary
+          deno-version: 2.0.6
```
crowlKats, line 26: done to prevent the upcoming deno doc changes to affect the reference API

philhawksworth:
I like it!

In future I think we can have some sort of combination of this logic, and a suitably populated 404 page with onward links to likely locations the user intended... but for now, this would be a valuable improvement

crowlKats:
our current 404 page is terrible (just an empty page that says "Not Found"), and since our URLs are structured hierarchically, going to the next possible parent makes the most sense.
One example also is for some reason google indexed https://docs.deno.com/api/node/streams, which doesnt even exist, so going to the parent in this case is a better experience

josh-collinsworth:
I still think overriding the "correct" response is hazardous and unexpected behavior. Even if it works sometimes, it probably won't others, and that creates a confusing experience. (I also don't think it's probably a good thing that we make it impossible to even hit a 404, for a few reasons.)

I also notice that if the first redirect doesn't hit an existing page, it creates a "too many redirects" error.

Can we write some kind of "is this what you were looking for?" logic into the 404 page that pattern matches, or something along those lines? That seems much more in line with expected behavior. I just think this seems like a pretty haphazard solution.

josh-collinsworth:
Happy to take a stab at this, if it helps

BlackAsLight:
I think it may lead users to be confused about what just happened if there isn't something on the page telling them they were redirected due to a 404. 

Like if I clicked a link and it just loaded the exact same page I'm on, I'd think this page is broken and not that the page I'm trying to go to doesn't exist. It would feel like a buggy experience.

crowlKats:
sure @josh-collinsworth, go ahead; i probably wont have have time to work on this anytime soon

.github/workflows/deploy.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Deno
         uses: denoland/setup-deno@v2
         with:
-          deno-version: canary
+          deno-version: 2.0.5
```
marvinhagemeister, line 26: Nit: Seems like an unrelated change

marvinhagemeister (approved):

PR #1166, opened 2024-11-20
Added install command for NextJS 14

greatvivek11:
The existing installation step doesn't work!! See https://questions.deno.com/m/1294573642491105343 https://github.com/denoland/deno/issues/26483

Some people may also want to migrate their existing Next13/14 apps to Deno2. This will help them clarify better instead of getting the latest Next setup with deno2.


kt3k:
This section seems like for starting a new Next.js project. I think migration topic should be covered in other pages.

(I also don't see what's the next step for migration after that command)

PR #1172, opened 2024-11-21
ci: check module graph of example scripts

kt3k:
This PR adds test cases which check the module graphs of the example scripts are valid.

This prevents the error like https://github.com/denoland/docs/pull/1153 in the future.


satyarohith (approved):
I like these automations to prevent similar bugs in the future.

LGTM!

Issue #1161, opened 2024-11-19
Title on docs pages is double encoded


marvinhagemeister:
![Screenshot 2024-11-19 at 16 38 18](https://github.com/user-attachments/assets/464b2504-e322-4680-96a0-e2c3c9228348)

It should not show raw html entities in the title bar.

PR #1173, opened 2024-11-21
replace encoded slash character in page titles

thisisjofrank:
Fixes https://github.com/denoland/docs/issues/1161

### DIFF
--- a/reference.page.jsx
+++ b/reference.page.jsx
@@ -27,9 +27,6 @@ export default function* () {
         title =
           match[1].slice(0, -"documentation".length).replace("&#x2F;", "/") +
           "- Deno Docs";
-        if (title.includes("/")) {
-          console.log(title);
-        }
       } catch (e) {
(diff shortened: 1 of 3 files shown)


reference.page.jsx
```diff
@@ -24,7 +24,9 @@ export default function* () {
       let title = "";
       try {
         const match = titleRegexp.exec(content);
-        title = match[1].slice(0, -"documentation".length) + "- Deno Docs";
+        title =
+          match[1].slice(0, -"documentation".length).replace("&#x2F;", "/") +
```
marvinhagemeister, line 28: The approach is correct. Let's use the `unescape()` function from here https://jsr.io/@std/html/doc/~/unescape so that we unescape all of them and don't have to revisit this in the future.

thisisjofrank: updated!

marvinhagemeister (approved):
Sweet!

PR #1177, opened 2024-11-21
use @db/sqlite example

boda-sh:
- use https://jsr.io/@db/sqlite instead of https://deno.land/x/sqlite3@0.12.0
- it is the same library 
- use the latest example from their README


kt3k (approved):
LGTM. Thank you!

Issue #1185, opened 2024-11-23
Why `Deno check` redirect to `Deno compile` 


forest1102:
CLI documentation for [Deno check](https://docs.deno.com/runtime/reference/cli/check/) redirects to [Deno compile](https://docs.deno.com/runtime/reference/cli/compile/)

PR #1186, opened 2024-11-23
fix: `deno check` redirecting to `deno compile`

marvinhagemeister:
Fixes https://github.com/denoland/docs/issues/1185


thisisjofrank (approved):
Oop good spot

PR #1189, opened 2024-11-24
doc: add a note about how to consume DenoRT

irbull:
Add documentation about how to consume DenoRT when building Deno from source.

Fixes: https://github.com/denoland/deno/issues/27035

### DIFF
--- a/runtime/contributing/building_from_source.md
+++ b/runtime/contributing/building_from_source.md
@@ -244,3 +244,19 @@ version instead of fetching it from `crates.io`.
 `Cargo.toml` match the version of the dependencies you have on disk.
 
 Use `cargo search <dependency_name>` to inspect the versions.
+
+## Consuming the DenoRT
+
+Deno Runtime (DenoRT) is an internal name given to the executable used to create
+a deno compile binary. It is a trimmed down version of the deno binary with the
+tooling and non-runtime stuff removed.
+
+By default, when `deno compile` is run, it will download a DenoRT that matches
+the version of the deno binary. If you want to use a custom DenoRT, you can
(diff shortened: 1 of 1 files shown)


runtime/contributing/building_from_source.md
```diff
@@ -244,3 +244,19 @@ version instead of fetching it from `crates.io`.
 `Cargo.toml` match the version of the dependencies you have on disk.
 
 Use `cargo search <dependency_name>` to inspect the versions.
+
+## Consuming the DenoRT
```
dsherret, line 248: Instead of documenting this, it might be nice for this to automatically happen when running `deno compile` from a `target/...` directory.

irbull: I've not seen the RFC for this, but this was in the comments of `get_base_binary`:
```
    // Phase 2 of the 'min sized' deno compile RFC talks
    // about adding this as a flag.
```
Having it automatically use the compiled `DenoRT` when the `path` includes `target/` seems like **The principle of least surprise**. That seems better than both the flag & env variable approach. I will often sym link my debug build into `/Users/irbull/.cargo/bin/deno`, and maybe we can also handle this case.

If you're interested, I could take a stab at this. I don't have too much time today to dive into it, but I could find some time this week.

dsherret: For sure!

irbull: I had a bit of time today between kids sports, so I brought my laptop to the rink and put this PR together: https://github.com/denoland/deno/pull/27052

I'll close this one.

PR #1192, opened 2024-11-25
[WIP]create pages for examples, tutorials and videos, refactor examples 

thisisjofrank:
(#1191)


_includes/raw_with_sidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+        />
+      </aside>
+      <div
+        class="absolute inset-0 backdrop-brightness-50 z-40 hidden sidebar-open:block sidebar-open:xl:hidden"
+        id="sidebar-cover"
+        data-open="false"
+      >
+      </div>
+      <div
+        class="absolute top-0 bottom-0 left-0 right-0 xl:left-74 overflow-y-auto xl:grid xl:grid-cols-7 xl:gap-8 max-w-screen-2xl mx-auto"
```
josh-collinsworth, line 106: Nit: `inset-0` sets top, bottom, left, and right all together.

_includes/raw_with_sidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+              viewBox="0 0 15 15"
+              width="16"
+              height="16"
+              class="text-foreground-secondary"
+            >
+              <g stroke="currentColor" stroke-width="1.2">
+                <path d="M.75.75l13.5 13.5M14.25.75L.75 14.25"></path>
+              </g>
+            </svg>
+          </button>
```
josh-collinsworth, line 90: This works for English, but it's better to use text, rather than an attribute, for localization/i18n purposes. Text is more easily translated, where attributes may not be. So the ideal markup here would look like this (overly simplified) example:

```html
<button>
  <svg aria-hidden="true" />
  <span class="sr-only">Close navigation bar</span>
</button>
```

learn/_components/CopyButton.tsx
```diff
@@ -0,0 +1,19 @@
+export function CopyButton(props: { text: string }) {
+  return (
+    <button
+      type="button"
+      aria-label="Copy code to clipboard"
```
josh-collinsworth, line 5: Same note about buttons and `.sr-only` text here; better to use text than `aria-label` (and to put `aria-hidden` on the SVG, just to be safe).

learn/_components/ExampleIcon.tsx
```diff
@@ -0,0 +1,19 @@
+export function ExampleIcon(props: { color: string }) {
+  return (
+    <svg
+      className="inline mr-2"
```
josh-collinsworth, line 4: Nit: I think it's better to style the SVG where it's used, rather than on the SVG itself. That way if we reuse the SVG elsewhere it doesn't come with a bunch of styling already in place to override. (Same note for the other two components.)

learn/_components/LearningList.tsx
```diff
@@ -0,0 +1,35 @@
...
+              className="homepage-link mb-1 runtime-link"
+              href={item.id}
+            >
+              {item.type === "tutorial" && <TutorialIcon color="#9d9d9d" />}
+              {item.type === "example" && <ExampleIcon color="#9d9d9d" />}
+              {item.type === "video" && <VideoIcon color="#9d9d9d" />}
+
+              {item.label}
+            </a>
+          </li>
```
josh-collinsworth, line 30: Two design notes on these list items:

1. In light mode, the link color doesn't provide enough contrast to pass a11y checks;
2. When viewing the whole big list, I feel like the icons kind of blend together. Would it make sense to differentiate the three categories by icon color, perhaps? Open to other ideas, and this doesn't need to block; just noticed that since every item has an icon, they visually struck me just as bullets and I ignored them until I looked closer.

learn/_components/SnippetComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+              props.firstOfFile ? "block" : "block sm:hidden"
+            }`}
+          >
+            {props.filename}
+          </span>
+        )}
+        <div class="-mx-4 h-full sm:mx-0 overflow-scroll sm:overflow-hidden relative gfm-highlight">
+          {props.snippet.code && (
+            <div
+              data-color-mode="light"
```
josh-collinsworth, line 35: Should this be `auto`?

learn/_pages/examples_test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+//     await t.step("Check graph: " + path, async () => {
+//       const result = await new Deno.Command(Deno.execPath(), {
+//         args: ["info", path],
+//       }).output();
+//       assertEquals(result.code, 0);
+//       assertNotMatch(decoder.decode(result.stdout), /\(resolve error\)/);
+//     });
+//   }
+// });
```
josh-collinsworth, line 21: Should this just be deleted?

learn/_pages/LandingPage.tsx
```diff
@@ -0,0 +1,105 @@
...
+              <input
+                type="checkbox"
+                id="example"
+                value="Examples"
+                className="mr-2"
+                checked
+              />
+              <span className="slider"></span>
+            </span>
+          </label>
```
josh-collinsworth, line 51: Issue: there's no accessible text for these labels, and no visible focus indication.

Also: ideally, these toggles should be grouped somehow, probably by a `<form>` element.

learn/_pages/LandingPage.tsx
```diff
@@ -0,0 +1,105 @@
...
+        {" "}
+        We welcome contributions!{" "}
+        <br />You can request more examples, or add your own at our{" "}
+        <a
+          href="https://github.com/denoland/deno-docs?tab=readme-ov-file#examples"
+          class="text-primary hover:underline focus:underline"
+        >
+          GitHub repository
+        </a>
+      </p>
```
josh-collinsworth, line 96: It's best to avoid `<br />` tags, as the user's font size, zoom, and/or screen size could make them appear in random, odd places.

Better solutions might be: using multiple paragraph tags; using `text-balance`; or setting a font-size-relative max-width.

learn_filter.client.ts
```diff
@@ -0,0 +1,24 @@
...
+      htmlItem.style.display = enabled ? "" : "none";
+    }
+  };
+
+  tutorial.addEventListener("change", () => filterItems());
+  example.addEventListener("change", () => filterItems());
+  video.addEventListener("change", () => filterItems());
+
+  filterItems();
+});
```
josh-collinsworth, line 24: FWIW, since we're just using radio inputs, we could probably achieve this in CSS using the `:has()` selector.

overrides.css
```diff
@@ -114,3 +114,61 @@
...
+  background-color: #2196f3;
+}
+
+input:focus + .slider {
+  box-shadow: 0 0 1px #2196f3;
+}
+
+input:checked + .slider:before {
+  transform: translateX(calc(0.8rem));
+}
```
josh-collinsworth, line 174: As a general rule, I think we should do everything we can in Tailwind, rather than maintaining two separate sources of truth for styling.

If we want to reuse styles (as with the `.switch` behavior, for example), then it's better to just create a reusable component with all the markup and classes, rather than have some styles in one place and some in another.

(I know this may be driven partially by dislike of Tailwind, which I can relate to if so, but I think it's better to stick with the system.)

josh-collinsworth (commented):
A lot of these comments are a bit nitpicky, but there are a few a11y issues I'd like to make sure we fix before launch.

Also: apologies if I jumped the gun on this review; I know it's still a draft.

_includes/raw_with_sidebar.tsx
thisisjofrank: I'm going to raise this as an issue rather than address it in this PR since this is the same sidebar that is used across the site.

learn/_pages/examples_test.ts
thisisjofrank: no, it needs to be fixed so that the tests actually pass

learn/_pages/LandingPage.tsx
thisisjofrank: Is the accessible text not the text in the label?

learn_filter.client.ts
thisisjofrank: I'll take a look

overrides.css
thisisjofrank: Yes, this was meant as a 'PoC' before John gets his hands on the design work. I wasn't sure if the toggle idea would be accepted so this was some throwaway until it is.

learn_filter.client.ts
thisisjofrank: I've moved this over to CSS, but will require some assistance optimising it for tailwind

PR #1198, opened 2024-11-27
fix: add `--unstable-sql` flag to formatter options

kotkifotki:
This pull request introduces the addition of the `--unstable-sql` flag, as introduced in https://github.com/denoland/deno/pull/26750, to the options of the formatter.

### DIFF
--- a/runtime/fundamentals/linting_and_formatting.md
+++ b/runtime/fundamentals/linting_and_formatting.md
@@ -111,6 +111,7 @@ before being merged.
 | unstable-css       | Enable formatting CSS, SCSS, Sass and Less files       |            |                         |
 | unstable-html      | Enable formatting HTML files                           |            |                         |
 | unstable-yaml      | Enable formatting YAML files                           |            |                         |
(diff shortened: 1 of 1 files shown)


kotkifotki:
Before proceeding with the merge, I would like to discuss whether the values used in the `deno.json` file should also be included, in addition to the flags. Currently, attempting to use these values results in a parsing error:

```bash
kotkifotki@laptop deno-project % deno fmt
error: Failed to parse "fmt" configuration

Caused by:
    unknown field `single-quote`, expected one of `useTabs`, `lineWidth`, `indentWidth`, `singleQuote`, `proseWrap`, `semiColons`, `options`, `include`, `exclude`, `files`
```

I would appreciate feedback on whether including these values is reasonable.

kt3k:
>  I would like to discuss whether the values used in the deno.json file should also be included, in addition to the flags.

There's a link below the table, which links to [this page](https://docs.deno.com/runtime/fundamentals/configuration/#formatting) that describes those deno.json fields.

kt3k (approved):
LGTM. Thanks for your contribution!

PR #1211, opened 2024-11-29
initial commit, adding header links to sidebar on mobile

thisisjofrank:


overrides.css
```diff
@@ -63,8 +63,8 @@ that are generated by deno_doc, the tool that powers the API reference documenta
   @apply ml-1 !important;
 }
 
-.doc .toc :not(.documentNavigation) > a {
-  @apply text-xl !important;
+.ddoc .z-50 {
+  z-index: 49 !important;
```
thisisjofrank, line 67: sorry for this hack, I can't edit the z-50 on the reference doc that is generated

philhawksworth:
This is great as it solves the immediate problem of the inability to navigate some parts of the top-level hierarchy on mobile, but I'm not quite sure about the behaviour in the API reference part of the site.

It feels of to me that the sub sections (deno, web, node) appear in the same format as additional top-level nav items in the sidebar when in that section.

If those secondary nav items appear in the sidebar, then I think they need to be visually distinct from the top-level nav items.

Screenshot 2024-12-11 at 17 26 03


It may be that the solution is also desirable anyway — to make the visual treatment of the top-level nav items in the sidebar a little different and contained from the other items. 

Astro have a good treatment of the same thing in their upcoming docs release that we might take some inspiration from:

Screenshot 2024-12-11 at 17 29 26

overrides.css
josh-collinsworth: Can you isolate it instead? May be a little cleaner. https://tailwindcss.com/docs/isolation

josh-collinsworth:
@philhawksworth @thisisjofrank I pushed a commit to stylize the "section picker" per Phil's feedback. I think it's in pretty ok shape, but I haven't tested it extensively. If you'd like to change it (or just blow the commit away) feel free.

thisisjofrank:
@josh-collinsworth I love this, looks super nice. We should apply it to the mobile sidebar on the homepage too, which somehow doesn't seem to be getting the styles. Can take a look tomorrow if you've no time.

philhawksworth (approved):
This is great! Going to make things feel far simpler

PR #1221, opened 2024-11-30
chore: prevent button text from wrapping

ryuapp:
Prevent text inside buttons from wrapping on iPhone SE.

Current:
image

This PR:
image


kt3k (approved):
LGTM. Thanks!

Issue #1088, opened 2024-10-29
Enhance Deno's Mocking Documentation and Examples


nirnay-raval-st10:
- Deno's current documentation, including the "Docs by Example" section, lacks comprehensive coverage of mocking techniques and practical examples. This makes it challenging for developers, especially those new to Deno, to understand and effectively utilize mocking in their testing strategies.

adit-bala:
Hey can i try making a pr and contributing this @thisisjofrank?

thisisjofrank:
> Hey can i try making a pr and contributing this @thisisjofrank?

Of course! That would be very appreciated.

adit-bala:
> > Hey can i try making a pr and contributing this @thisisjofrank?
> 
> Of course! That would be very appreciated.

Will try and get something in by the end of the week!

PR #1222, opened 2024-11-30
docs(examples): add example for spy functions

adairo:
Hello deno community!
I added an example for how to spy on functions using `std@testing/mock`

Related to issue #1088 

I wasn't sure in which group put this example so I kinda created a new group for 'Testing'. I want to upload another example on how to use stubs. It seems like the group should be created elsewhere in order to appear as a correct group (titlecase), I would really appreciate little help on this :)

### DIFF
--- a/examples/spy-functions.ts
+++ b/examples/spy-functions.ts
@@ -14,36 +14,36 @@
 import { assertSpyCall, assertSpyCalls, spy } from "jsr:@std/testing/mock";
 
 function log(message: string) {
-    console.log(message);
+  console.log(message);
 }
 
 const logger = { log };
 
 Deno.test("logger uses the log function", () => {
-    // Create a spy on the `logger.log` method.
-    const logSpy = spy(logger, "log");
+  // Create a spy on the `logger.log` method.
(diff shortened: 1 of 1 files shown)


examples/spy-functions.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * @title Spy functions
+ * @difficulty intermediate
+ * @tags cli, deploy
+ * @run <url>
+ * @resource {https://jsr.io/@std/testing/doc/mock#spying} Spy docs on JSR
+ * @resource {https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html} Typescript docs for `using` keyword
```
kt3k, line 7: Good reference 👍

kt3k (approved):
LGTM

(Maybe we can also add more advanced examples where spied functions are called indirectly via some other functions. But this one seems just fine for giving the basic idea of spying)

adairo:
Thank you very much @kt3k for taking the time to review this. For sure I'll consider adding more advanced examples.

PR #1231, opened 2024-12-03
docs(example): add new example for test stubs

adairo:
Hi everyone!

Here is an example of how to use test stubs on Deno.

related to #1088

Hope it helps new deno users to get a grasp of these cool testing features.


kt3k (approved):
LGTM. Thanks!

PR #1232, opened 2024-12-04
Add DuckDB Example

YTKme:
Adding an example for using DuckDB.

Few items to note as I was working on the example

- I was working with Windows 11 and WSL 2 on Debian, and followed the [Pre-requisite setup (Linux)](https://github.com/suketa/ruby-duckdb?tab=readme-ov-file#pre-requisite-setup-linux) steps from [suketa/ruby-duckdb](https://github.com/suketa/ruby-duckdb) to install DuckDB to resolve `Error: Could not open library: Could not open library: libduckdb.so`.
- Encountered issue with `query` function when using [littledivy/duckdb](https://github.com/littledivy/duckdb) binding from [deno.land/x](https://deno.land/x/duckdb), but `stream` and `prepare` example worked fine. (will try to file issue for this one)


kt3k:
I'm not sure duckdb is worth mentioning here... Is it comparable with other entries like `redis`, `postgres`, `supabase`, `mongodb`, or `sqlite`?

YTKme:
> I'm not sure duckdb is worth mentioning here... Is it comparable with other entries like `redis`, `postgres`, `supabase`, `mongodb`, or `sqlite`?

There are similarities. The current use cases are mainly for data science, and can handle concurrency well. It basically is a combination of `redis` and `sqlite`, and I believe there would be some use cases for it.

kt3k (approved):
Ok. Let's try to include this

LGTM

(Note: x/duckdb is Divy's project)

PR #1233, opened 2024-12-04
Temporary fix: Search with google

philhawksworth:
To help us assess alternatives for a better search experience for visitors with more complete and accurate results, we could direct search requests to a scoped search query on google.

This PR:
- Replaces the Orama search UI with a new search input which directs the query to Google's search
- Removes all Orama CSS, JS, and UI components
- Removes the Orama search index generation tasks from the build
- Removes any other Orama dependencies

Ideally we'll reinstate Orama once this is optimised suitably and includes all the search results we would expect. Or potentially an alternative service


ry (commented):
This is approved from a strategic perspective, would like @josh-collinsworth's review on code

_components/SearchInput.tsx
```diff
@@ -0,0 +1,31 @@
+export default function SearchInput() {
+  return (
+    <form action="https://www.google.co.uk/search" method="get" class="search">
```
lucacasonato, line 3:
```suggestion
    <form action="https://www.google.com/search" method="get" class="search">
```

uncenter:
Is there a tracking issue for future search improvements?

PR #1237, opened 2024-12-06
docs(security): mention --alllow-env wildcards

dandv:
I'm surprised this wasn't in the docs already. Is there a more in-depth treatment of the `--allow-env` flag somewhere else in the docs (not in [the blog post](https://deno.com/blog/v2.1#--allow-env-wildcards))?

### DIFF
--- a/runtime/fundamentals/security.md
+++ b/runtime/fundamentals/security.md
@@ -246,7 +246,8 @@ includes reading environment variables, and setting new values.
 
 Access to environment variables is granted using the `--allow-env` flag. This
 flag can be specified with a list of environment variables to allow access to
-specific environment variables.
+specific environment variables. Starting with Deno v2.1, you can now specify
+suffix wildcards to allow “scoped” access to environmental variables.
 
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/security.md
```diff
@@ -256,8 +257,11 @@ deno run -E script.ts
 # or
 deno run --allow-env script.ts
 
-# Allow HOME and FOO environment variable
+# Allow HOME and FOO environment variables
 deno run --allow-env=HOME,FOO script.ts
+
+# Allow access to all environment variables starting with AWS_
+deno run --allow-env=AWS_* script.ts
```
kt3k, line 264: Probably it's better to surround the value with quotes as some shells try to expand `*` and fail:

```suggestion
deno run --allow-env="AWS_*" script.ts
```

kt3k (approved):
LGTM

runtime/fundamentals/security.md
dandv: The blog post doesn't quote them...

PR #1246, opened 2024-12-10
fix: type safety for yaml parsed file

securesvet:
`parsed` is now casted to `DescriptionItem` instead of being `unknown`

### DIFF
--- a/runtime/_data.ts
+++ b/runtime/_data.ts
@@ -292,7 +292,7 @@ export async function generateDescriptions(): Promise<Descriptions> {
     )
   ) {
     const file = await Deno.readTextFile(dirEntry.path);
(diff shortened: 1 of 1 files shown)


runtime/_data.ts
```diff
@@ -292,7 +292,7 @@ export async function generateDescriptions(): Promise<Descriptions> {
     )
   ) {
     const file = await Deno.readTextFile(dirEntry.path);
-    const parsed = yamlParse(file);
+    const parsed = yamlParse(file) as DescriptionItem;
```
kt3k, line 295: The code after here is for validating that `parsed` is in the form of `DescriptionItem`. It doesn't seem correct to cast this type to `DescriptionItem` here

PR #1259, opened 2024-12-12
adding the main navigation to the mobile nav sidebar, 

thisisjofrank:
adding a totop button to the API ref docs, adding the mobile nav to the API docs


philhawksworth (approved):
LGTM!

I see one slight issue with the sidebar scrolling away in the APIs sections. But I think we'd be better shipping this and addressing that as a fast follow than not having the ability to navigate around on mobile.

PR #1263, opened 2024-12-13
Examples redirects

thisisjofrank:
add redirects for renamed files and update dashes to underscores


philhawksworth (approved):
I think that's got it. Nice one!

PR #1268, opened 2024-12-15
docs: update benchmark file reading example

raisiqueira:
The current example in the benchmark documentation about _file reading_ breaks in the `file.close()` line. Also, I updated to include an example below of the _file reading_ bench saying that this benchmark needs the `allow-read` flag.

CleanShot 2024-12-15 at 15 52 39@2x

### DIFF
--- a/runtime/reference/cli/bench.md
+++ b/runtime/reference/cli/bench.md
@@ -117,7 +117,8 @@ Deno.bench("foo", async (b) => {
 });
 ```
 
-The above example requires the `--allow-read` flag to run the benchmark: `deno bench file_reading.ts --allow-read`.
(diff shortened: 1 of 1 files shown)


runtime/reference/cli/bench.md
```diff
@@ -114,13 +114,12 @@ Deno.bench("foo", async (b) => {
 
   // End measurement here.
   b.end();
-
-  // Now we can perform some potentially time-consuming teardown that will not
-  // taint out benchmark results.
-  file.close();
```
kt3k, line 120: Probably this becomes an error at some point when we migrated from Reader/Writer pattern to Web Stream API in the above code (The above web stream call closes the file automatically when it's read)

PR #1277, opened 2024-12-17
refactor: make lint rule docs better looking

crowlKats:


josh-collinsworth:
Is there a reason we have `jsx` and `react` in here? I don't see any that currently use those rules. Guessing for future work?

In any case, we should probably flesh out the descriptions up top a little bit more.

bartlomieju:
@josh-collinsworth yes, we got a bunch of incoming rules that have these tags that will be available once https://github.com/denoland/deno/pull/27162 lands.

bartlomieju (approved):
LGTM, going to land it as is and further improve in follow up PRs.

PR #1278, opened 2024-12-17
perf optimisations (5~7 seconds shaved off (windows))

thisisjofrank:
1. Remove cleaning - saves 1.5 seconds per update
2. Remove parsing all .html files to add highlighting and tabs. - highlighting configured in markdown plugin, tabs executed on client side. Additionally server side highlighting in snippet component.
3. Configured scoped updates and debounce for responsiveness.
4. Added better logging using Lume's logger.
5. Create secondary serve task which sets default log level to WARN and disables link checker plugin in dev-mode using an environment variable as it was taking over half the cycle time (5s+)

### DIFF
--- a/_config.ts
+++ b/_config.ts
@@ -101,6 +101,7 @@ const site = lume(
 );
 
 site.copy("static", ".");
+site.copy("timeUtils.ts");
 site.copy("subhosting/api/images");
 site.copy("deploy/docs-images");
 site.copy("deploy/kv/manual/images");
--- a/_config.ts
+++ b/_config.ts
@@ -55,6 +55,7 @@ const site = lume(
         "/.github",
         "/.vscode",
       ],
+      debounce: 1_000,
     },
   },
   {
--- a/_config.ts
+++ b/_config.ts
@@ -186,7 +186,7 @@ const SKIP_CHECK_URLS = (Deno.env.get("SKIP_CHECK_URLS") || "false")
   .toLowerCase();
 
 if (SKIP_CHECK_URLS !== "true") {
-  log.info(`${cliNow()} Enabling broken link checker`);
+  log.info(`${cliNow()} Enabling broken-link checker`);
 
   site.use(
     checkUrls({
--- a/_config.ts
+++ b/_config.ts
@@ -30,6 +30,8 @@ import createGAMiddleware from "./middleware/googleAnalytics.ts";
 import redirectsMiddleware, {
   toFileAndInMemory,
 } from "./middleware/redirects.ts";
+import { cliNow } from "./timeUtils.ts";
+import { log } from "lume/core/utils/log.ts";
 
 const site = lume(
   {
@@ -53,7 +55,6 @@ const site = lume(
         "/.github",
         "/.vscode",
       ],
-      debounce: 1_000,
     },
   },
   {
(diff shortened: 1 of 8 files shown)


_config.ts
```diff
@@ -125,28 +143,13 @@ site.use(
...
-    const document = page.document!;
-    if (!document.querySelector(".ddoc")) {
-      document.body.classList.add("apply-prism");
-      document
-        .querySelectorAll("body.apply-prism pre code")
-        .forEach((element) => Prism.highlightElement(element));
-    }
-  }
-});
-
```
marvinhagemeister, line 143: Great to see this being removed! Does the highlighting still work on reference pages? I could be wrong but iirc it was used there to add highlighting at some point.

_config.ts
```diff
@@ -161,44 +164,6 @@ site.copy("reference_gen/gen/node/page.css", "/api/node/page.css");
 site.copy("reference_gen/gen/node/styles.css", "/api/node/styles.css");
 site.copy("reference_gen/gen/node/script.js", "/api/node/script.js");
 
-site.process([".html"], (pages) => {
-  for (const page of pages) {
-    const document = page.document;
-    if (document) {
-      const tabGroups = document.querySelectorAll("deno-tabs");
```
marvinhagemeister, line 168: I'm all for removing this! Putting it in the client js works, but might incur an additional layout shift. That said if this makes working on the docs much more pleasent, I'm all for it.

marvinhagemeister (commented):
great finds!

_config.ts
thisisjofrank: Yeah, there's a teeny tiny, split second shift, but I'm not sure if even notice it if I wasn't looking for it...

_config.ts
thisisjofrank: As far as I can see, the highlighting is still working in API ref. I do notice the styles are slightly different between local and prod, but I think that might be the same on main... :S

philhawksworth (approved):
Fantastic

philhawksworth (approved):
good call

_config.ts
```diff
@@ -210,13 +176,35 @@ site.ignore(
...
+  (path) => path == "/overrides.css",
+  (path) => /\.(js|ts)$/.test(path),
+  (path) => path.startsWith("/api/deno/"),
 );
 
+const SKIP_CHECK_URLS = (Deno.env.get("SKIP_CHECK_URLS") || "false")
+  .toLowerCase();
+
+if (SKIP_CHECK_URLS !== "true") {
+  log.info(`${cliNow()} Enabling broken link checker`);
```
marvinhagemeister, line 189: Nit: wording sounds like the link checker itself is broken

_config.ts
```diff
@@ -210,13 +176,35 @@ site.ignore(
...
+site.scopedUpdates(
+  (path) => path == "/overrides.css",
+  (path) => /\.(js|ts)$/.test(path),
+  (path) => path.startsWith("/api/deno/"),
 );
 
+const SKIP_CHECK_URLS = (Deno.env.get("SKIP_CHECK_URLS") || "false")
+  .toLowerCase();
+
+if (SKIP_CHECK_URLS !== "true") {
```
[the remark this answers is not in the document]
thisisjofrank: good point, will add env variable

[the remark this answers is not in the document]
thisisjofrank: Wait, if there is no setting it will default to "don't skip", so it will still run if you use deno task serve, and it will run on CI
My thought being that on deploy the extra 5 seconds that the link checker takes probably are fine... (maybe)

_config.ts
thisisjofrank: added a dash, so now it is "Enabling broken-link checker"

PR #1281, opened 2024-12-19
feat: add modified timestamp

### LINKED ISSUE #1280: Add last modified timestmp to content pages in docs
A good suggestion from the community is for us to include a timestamp on the docs to indicate when a page was last updated.

There will be som nuance for us to consider here as some pages are autogenerated from source files while others are more explicitly authored.

Consider using the file's last modified timestamp and displaying the date in a suitable location in the page template.

crowlKats:
Closes #1280

### DIFF
--- a/_components/Feedback.tsx
+++ b/_components/Feedback.tsx
@@ -3,6 +3,20 @@ export default function Feedback({ file }) {
     return <></>;
   } else {
     const githubPath = "https://github.com/denoland/docs/edit/main" + file;
+
+    let modifiedDate = null;
+    try {
+      const dateString = new TextDecoder().decode((new Deno.Command("git", {
+        args: [
+          "log", "-1", "--pretty=%cI", "./" + file
+        ],
+      })).outputSync().stdout);
+
+      modifiedDate = new Date(dateString);
+    } catch (e) {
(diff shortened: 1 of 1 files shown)


_components/Feedback.tsx
```diff
@@ -138,6 +157,20 @@ export default function Feedback({ file }: { file: string | undefined }) {
                   </button>
                 </div>
               </div>
+              {modifiedDate &&
+                (
+                  <div class="mt-4">
+                    This page was last modified on{" "}
+                    <time datetime={modifiedDate.toString()}>
```
scarf005, line 164:
```suggestion
                    <time datetime={modifiedDate.toISOString()}>
```

> The toString() method of [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) instances returns a string representing this date interpreted in the local timezone.

[`toString()` returns localized date;](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString) but `<time>` requires [datetime in ISO8601 format](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values).

PR #1294, opened 2024-12-27
Fix typos around parallel dependency execution in task.md

dandv:


runtime/reference/cli/task.md
```diff
@@ -109,9 +109,9 @@ Task serve deno run -RN server.ts
 Listening on http://localhost:8000/
 ```
 
-Dependency tasks are in parallel, with the default parallel limit being equal to
-number of cores on your machine. To change this limit use `DENO_JOBS`
-environmental variable.
+Dependency tasks are in executed parallel, with the default parallel limit being
+equal to the number of cores on your machine. To change this limit use the
+`DENO_JOBS` environmental variable.
```
thisisjofrank, line 114:
```suggestion
Dependency tasks are executed in parallel, with the default parallel limit being
equal to number of cores on your machine. To change this limit, use the
`DENO_JOBS` environmental variable.
```

thisisjofrank (approved):

PR #1301, opened 2025-01-01
Update helix LSP config

senekor:
It is not necessary to override all fields of the language configuration table. JavaScript needs to be configured in a separate table.


marvinhagemeister (approved):
Thanks!

PR #1304, opened 2025-01-02
remove max-width on feedback and page nav

donjo:
we've expanded the max width on the rest of the content so it probably makes sense to let the feedback module and page navigation expand to the same width so they aren't oddly centered under some wider content.

Old:
Screenshot 2025-01-02 at 4 28 17 PM

New:
Screenshot 2025-01-02 at 4 28 50 PM


thisisjofrank (approved):
Oh this looks so much better!

PR #1306, opened 2025-01-04
Add note the difference between `deno.json` and `--import-map` option

lambdalisue:
See https://github.com/denoland/deno/issues/27539

# Rendering result

CleanShot 2025-01-04 at 12 04 04


thisisjofrank (approved):
This is a great addition ,thank you.

PR #1319, opened 2025-01-08
Expand FFI in security.md

dandv:
Should also match the anchor at https://github.com/denoland/docs/blob/main/go.json#L4


kt3k (approved):
Good catch. Thanks!

PR #1322, opened 2025-01-09
Update to correct url for deno run hello world example in command_line_interface.md

MxMossy:
Updated the url to the the correct https://docs.deno.com/examples/scripts/hello_world.ts .
Previous url led to a module not found error when using deno run.


thisisjofrank (approved):
Good catch, thank you!

PR #1323, opened 2025-01-09
Edit language to remove reference to future date, which is now past

philhawksworth:
We refer to the release date of v2.1 as if it were in the future, but that date has now past.
This edit removes the reference to the date being in the future on this page:

https://docs.deno.com/runtime/fundamentals/stability_and_releases/#long-term-support-(lts)


marvinhagemeister (approved):
LGTM, probably needs a `deno fmt`

PR #1325, opened 2025-01-10
Nuxt tutorial

lambtron:
Added a new tutorial!


examples/tutorials/nuxt.md
```diff
@@ -0,0 +1,423 @@
+---
+title: "Build a Nuxt app with Deno"
+url: /examples/nuxt_tutorial/
+---
+
+[Nuxt](https://nuxt.com/) is a framework that provides an intuitive way to
+create full-stack applications based on [Vue](https://vuejs.org/). It offers
+file-based routing, server-side rendering capabilities, and automatic code
```
philhawksworth, line 8: One of the benefits of Nuxt is how you can mix lots of different rendering models. So I suggest this small edit here:

```suggestion
file-based routing, a variety of rendering options, and automatic code
```

examples/tutorials/nuxt.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Scaffold a Nuxt app with Deno
+
+We can create a new Nuxt project using deno like this:
+
+```bash
+deno -A npm:nuxi@latest init
+```
+
+Here, we’re using npm as the package manager. This will create a nuxt-app with
```
philhawksworth, line 34: Is that the best language when talking about the package manager? Might it be more correct to say that we're using Deno as the package manager, and using the package from NPM?

These might be silly semantics, but I'm thinking of landing the point over time, that deno is a package manager too.

So, perhaps...

```suggestion
We'll use Deno to manage our package dependencies, and can grab the Nuxt package from npm. This will create a nuxt-app with
```

examples/tutorials/nuxt.md
```diff
@@ -0,0 +1,423 @@
...
+// assets/css/main.css
+
+base;
+components;
+utilities;
+```
+
+The only other thing we'll need to do is update our `nuxt.config.ts` file to
+configure our Nuxt application for Deno compatibility, enable development tools,
+and set up Tailwind CSS. We're disabling SSR for better Deno compatibility and
```
philhawksworth, line 356: Is there a way to phrase this that implies less of a limitation for Deno? It sounds a but like Deno can't do SSR. It might be a touch ambiguous if we mean the runtime or Deno Deploy when we say that too.

Maybe...

```suggestion
and set up Tailwind CSS. We're disabling SSR and instead generating the entire site for the greatest flexibility in deployment options, and
```

philhawksworth (approved):
Looks great!
I have a few non-blocking suggestions for your consideration.

Lexpeartha:
@lambtron have you noticed that "Next Steps" section recommends module that is Nuxt v2 only?

Issue #1292, opened 2024-12-26
Code copying is not working under tabs


ZYSzys:
![image](https://github.com/user-attachments/assets/c9aa0017-c490-41f3-a9f2-215b3b3f8f72)

This bug was introduced in https://github.com/denoland/docs/pull/1278

TriMoon:
Mentioned merge, seems to have fixed it https://github.com/denoland/docs/pull/1326#pullrequestreview-2563645691.
So if it has been fixed for you too, please close this issue 😉

thisisjofrank:
This  now seems to be fixed

PR #1326, opened 2025-01-12
fix: restore copy button in tabs

tolluset:
ref: https://github.com/denoland/docs/issues/1292

I fixed the copy button in the tabs.

The reason the copy button in the tabs didn't work was because of the render order issue. So I moved the binding delegate to the document.

tested in local: 

https://github.com/user-attachments/assets/b016d81d-b25b-4821-a7ca-fd8e9c53de64

### DIFF
--- a/copy.client.ts
+++ b/copy.client.ts
@@ -1,10 +1,31 @@
-const copyBtns = document.querySelectorAll("button[data-copy]");
+document.addEventListener("click", (event) => {
+  const btn = (event.target as HTMLElement).closest("button[data-copy]");
 
-copyBtns.forEach((btn) => {
-  btn.addEventListener("click", () => {
-    let textToCopy = btn.getAttribute("data-copy") as string;
-    // CLEAN COMMANDS:  Remove leading spaces, $, and > from each line
-    textToCopy = textToCopy.replace(/^[\$>\s]+/, "");
-    navigator?.clipboard?.writeText(textToCopy);
+  if (!btn) {
+    return;
+  }
+
+  let textToCopy = btn.getAttribute("data-copy") as string;
+
+  // CLEAN COMMANDS:  Remove leading spaces, $, and > from each line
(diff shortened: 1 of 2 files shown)


copy.client.ts
```diff
@@ -1,10 +1,31 @@
-const copyBtns = document.querySelectorAll("button[data-copy]");
-
-copyBtns.forEach((btn) => {
-  btn.addEventListener("click", () => {
-    let textToCopy = btn.getAttribute("data-copy") as string;
-    // CLEAN COMMANDS:  Remove leading spaces, $, and > from each line
-    textToCopy = textToCopy.replace(/^[\$>\s]+/, "");
-    navigator?.clipboard?.writeText(textToCopy);
+document.addEventListener("click", (event) => {
+  const btn = (event.target as HTMLElement).closest("button[data-copy]");
```
tolluset, line 2: The button has svg icons, so use `closest` to prevent missing copies.

markdown-it/codeblock-copy.ts
```diff
@@ -30,28 +30,6 @@ export default function codeblockCopyPlugin(md: any) {
       </button>
     `;
 
-    const script = `
```
tolluset, line 33: Moved to the `copy.client.ts` file.

copy.client.ts
```diff
@@ -1,10 +1,31 @@
...
+
+  let textToCopy = btn.getAttribute("data-copy") as string;
+
+  // CLEAN COMMANDS:  Remove leading spaces, $, and > from each line
+  textToCopy = textToCopy.replace(/^[\$>\s]+/, "");
+
+  navigator?.clipboard?.writeText(textToCopy).then(() => {
+    if (!btn) {
+      return;
+    }
```
kt3k, line 16: nit: I don't think this can be null or undefined even if the button was removed at this point

kt3k (approved):
Confirmed this fixes the issue locally. Thanks for the important fix!

PR #1329, opened 2025-01-13
fix(examples): a few changes to nuxt example to make it more idiomatic

danielroe:
I would also suggest using https://tailwindcss.nuxtjs.org/ which should be as simple to set up as running `deno -A npm:nuxi@latest module add tailwindcss` and requires no additional configuration.

you can see my comments on the specific code changes [here](https://github.com/denoland/examples/pull/33)


philhawksworth (approved):
Great! Thanks for these improvements, @danielroe!

PR #1337, opened 2025-01-16
Display last updated date

philhawksworth:
- Adds date to content pages
- Adds conditional operations to Lume for expensive tasks, suitable for skipping during dev work.  
Opt in with `BUILD_TYPE=FULL` env var when running build task
- `BUILD_TYPE=FULL` flag added to production build task
- `deno task serve:full` task added for developing with inclusion of expensive operations
- Production and full builds enhance the date from the file creation date to the git last modified date

### DIFF
--- a/_includes/doc.tsx
+++ b/_includes/doc.tsx
@@ -19,6 +19,14 @@ export default function Page(props: Lume.Data, helpers: Lume.Helpers) {
     throw new Error("Missing sidebar for " + props.url);
   }
 
+  function displayDate(date: string) {
+    return new Date(date).toLocaleDateString("en-US", {
+      year: "numeric",
+      month: "short",
+      day: "numeric",
+    });
+  }
+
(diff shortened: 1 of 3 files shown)


_includes/doc.tsx
```diff
@@ -19,6 +19,14 @@ export default function Page(props: Lume.Data, helpers: Lume.Helpers) {
     throw new Error("Missing sidebar for " + props.url);
   }
 
+  function displayDate(date: string) {
+    return new Date(date).toLocaleDateString("en-US", {
+      year: "numeric",
+      month: "short",
+      day: "numeric",
+    });
+  }
```
ry, line 28: I'd prefer to render this as YYYY-mm-dd

also I think this should take a `Date` arg rather than string

```ts
function displayDate(date: Date): string {
  return date.toISOString().slice(0, 10);
}
```

PR #1346, opened 2025-01-20
fix: use npm library in jwt example

### LINKED ISSUE #1345: Update JWT example
https://x.com/meathill1/status/1880860822844678210

Should use JSR and/or NPM

ryuapp:
closes #1345 
It seems that jose is not published in jsr for the following reasons: https://github.com/panva/jose/discussions/695

https://docs.deno.com/examples/creating_and_verifying_jwt/


kt3k (approved):
LGTM

PR #1363, opened 2025-01-24
feat: alternative docs gen

crowlKats:


thisisjofrank (approved):
This looks great to me, a good jumping off point for fixing up the html/css.

PR #1365, opened 2025-01-24
fix: Update task.md to fix error with package name

fro-profesional:
image

For deno workspaces name is expected to be in the tasks? or should it be a the top level in the deno.json?


marvinhagemeister (approved):
Good catch, thanks for fixing this. The name should indeed be at the top 👍

PR #1370, opened 2025-01-27
Minor style fix

thisisjofrank:


philhawksworth (approved):
Looks good.

I notice there is a similar issue on another part of the site which might be worth rolling into this PR too. 

Take a look at 
`/reference/_components/DocBlockSubtitleClass.tsx` which has a similar issue  for both extends` and `implements`

PR #1374, opened 2025-01-28
fix: use server-side GA on more requests

ry:
untested - the theory is that server-side GA isn't catching a lot of paths.


crowlKats (approved):
LGTM

PR #1380, opened 2025-01-30
Replace https import with the JSR example

philhawksworth:
Update to use JSR and remove https import example

### DIFF
--- a/examples/scripts/import_export.ts
+++ b/examples/scripts/import_export.ts
@@ -39,5 +39,5 @@ util.sayHello("World");
 import { camelCase } from "jsr:@luca/cases@1";
 console.log(camelCase("hello world")); // helloWorld
 
-import OpenAI from jsr:@openai/openai 
+import OpenAI from "jsr:@openai/openai"
 const client = new OpenAI();
--- a/examples/scripts/import_export.ts
+++ b/examples/scripts/import_export.ts
@@ -34,10 +34,10 @@ sayHello("World");
 import * as util from "./util.ts";
 util.sayHello("World");
 
-// Imports don't have to be relative, they can also reference absolute file,
-// https, or [JSR](https://jsr.io) URLs.
+// Imports don't have to be relative, they can also reference 
+// absolute file, npm, or [JSR](https://jsr.io) URLs.
 import { camelCase } from "jsr:@luca/cases@1";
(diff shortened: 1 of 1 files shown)


examples/scripts/import_export.ts
```diff
@@ -34,10 +34,10 @@ sayHello("World");
...
 
-// Imports don't have to be relative, they can also reference absolute file,
-// https, or [JSR](https://jsr.io) URLs.
+// Imports don't have to be relative, they can also reference 
+// absolute file, npm, or [JSR](https://jsr.io) URLs.
 import { camelCase } from "jsr:@luca/cases@1";
 console.log(camelCase("hello world")); // helloWorld
 
-import OpenAI from "https://deno.land/x/openai@v4.53.0/mod.ts";
+import OpenAI from jsr:@openai/openai
```
Blocksnmore, line 42:
```suggestion
import OpenAI from "jsr:@openai/openai"
```

examples/scripts/import_export.ts
```diff
@@ -34,10 +34,10 @@ sayHello("World");
 import * as util from "./util.ts";
 util.sayHello("World");
 
-// Imports don't have to be relative, they can also reference absolute file,
-// https, or [JSR](https://jsr.io) URLs.
+// Imports don't have to be relative, they can also reference 
+// absolute file, npm, or [JSR](https://jsr.io) URLs.
```
Blocksnmore, line 38: I’d argue keeping the https mentioned as it’s a feature in deno, just because deno.land/x is replaced with jsr doesn’t mean you still can’t use http imports for say esm.sh

Blocksnmore (commented):
Just some things I noticed, this pr should also probably be merged with #1379 rather than keeping them separate

examples/scripts/import_export.ts
philhawksworth: Fair. Agreed.

philhawksworth:
> Just some things I noticed, this pr should also probably be merged with #1379 rather than keeping them separate

I'll merge these as they are... but there are more to come like this, so I'll batch them into reasonably-sized PRs. I don't want any single PR to get too large to very easily review and merge, but agree that we can be less granular than this.

marvinhagemeister (approved):

PR #1388, opened 2025-01-31
test

ry:


bartlomieju:
This will pass CI because of https://github.com/denoland/docs/pull/1385. You'll need to change the version of Deno used to 2.1.9 to see if this works

server.ts
```diff
@@ -1,3 +1,5 @@
+process.env = Deno.env.toObject();
```
bartlomieju, line 1:
```suggestion
// process.env = Deno.env.toObject();
import "data:application/javascript;charset=utf-8;base64,cHJvY2Vzcy5lbnYgPSBEZW5vLmVudi50b09iamVjdCgpOw=="
```
Otherwise the imports will be evaluated before your code to assing to process

PR #1397, opened 2025-02-05
feat(lint): add page for custom lint plugins

marvinhagemeister:
This PR adds a page about using and creating custom lint plugins. This is a feature that will be released with the next version  Deno `2.2.0`.


philhawksworth (approved):
Looks good!
I think it needs a quick `deno fmt` and we should hold from merging until v2.2 lands.

PR #1400, opened 2025-02-06
remove duplicate url checker

thisisjofrank:


philhawksworth (approved):
Yep! We can only really rely on the checker in the CI.
LGTM

PR #1403, opened 2025-02-06
Feedback to notion

philhawksworth:
- Updates feedback middleware to send the feedback to Notion rather than Google sheets


middleware/functions/feedback-notion.ts
```diff
@@ -0,0 +1,109 @@
+import { Client } from "npm:@notionhq/client";
+import type { FeedbackSubmission } from "../../types.ts";
+
+// Collecting entries to this internal Notion page:
+// https://www.notion.so/denolandinc/192771466362800da55dc4552ad9a70c?v=192771466362803793d5000c0f522280
```
thisisjofrank, line 5: do we want to publicise this?

philhawksworth: It's private so I thought there was no harm and it was convenient... but I'm happy to remove it if that's our preference

thisisjofrank: I think probably better not shared just to discourage eyes from trying to look at it.

philhawksworth: I've removed it

philhawksworth:
Test submissions work as expected

thisisjofrank (approved):
woop!

PR #1407, opened 2025-02-08
fix(lint): rename `no-process-global` rule

petamoriken:
ref: https://github.com/denoland/deno/issues/27998


thisisjofrank (approved):
This looks good to me. thank you!

PR #1411, opened 2025-02-11
:memo: Prisma: Updare `--unstable` flag usage

homersimpsons:
The `--unstable` flag was removed in Deno 2.0. Replace it with the more precise `--unstable-detect-cjs` flag.


marvinhagemeister (approved):
Thanks!

PR #1416, opened 2025-02-11
doc: add multiple package entries example

marvinhagemeister:
This adds a section on how to declare multiple package entries.

I feel like long term we should have a dedicated "packaging" page. Right now the only information about how to package your code is here on the workspace page.

Fixes https://github.com/denoland/docs/issues/1408


philhawksworth (approved):
LGTM!

PR #1424, opened 2025-02-13
Update examples: file exists, folder exists

philhawksworth:
- Updates the language and details of the example foe checking a file exists
- Adds an example for checking a folder exists


thisisjofrank (approved):
👍

PR #1437, opened 2025-02-18
Update node:sqlite compatibility

bartlomieju:


ry (approved):
Mention that this was added in Deno 2.2 - otherwise lgtm

PR #1461, opened 2025-02-20
Update release cadence language

philhawksworth:
Update the information on deno minor release cadence to reflect the 12 weeks timetable.

Effected page: `/runtime/contributing/release_schedule/`


marvinhagemeister (approved):
LGTM, we should probably also update the blogpost where we're showing the timeline of monthly releases.

philhawksworth:
> LGTM, we should probably also update the blogpost where we're showing the timeline of monthly releases.

Oh good call!

PR #1462, opened 2025-02-20
Fix overflowing content

josh-collinsworth:
Fixes the horizontal overflow issue on some pages by explicitly setting column minimum widths to 0, and by removing minimum width from the sidebars. This gives the grid full control over the column widths, and doesn't allow any grid items to force the columns to be wider.

CleanShot 2025-02-20 at 13 52 47@2x

### DIFF
--- a/_components/Navigation.css
+++ b/_components/Navigation.css
@@ -1,6 +1,5 @@
 .nav {
   height: 100%;
-  min-width: 16rem;
   border-right: 1px solid hsl(var(--foreground-tertiary));
   font-size: 0.8rem;
   background-color: hsl(var(--background-primary));
--- a/_components/Navigation.css
+++ b/_components/Navigation.css
@@ -1,6 +1,6 @@
 .nav {
   height: 100%;
-  min-width: 280px;
+  min-width: 16rem;
   border-right: 1px solid hsl(var(--foreground-tertiary));
   font-size: 0.8rem;
   background-color: hsl(var(--background-primary));
--- a/_components/TableOfContents.css
+++ b/_components/TableOfContents.css
@@ -9,7 +9,7 @@
 @media (min-width: 1024px) {
   .content {
     display: grid;
-    grid-template-columns: minmax(auto, 960px) 1fr;
+    grid-template-columns: minmax(0, 8fr) minmax(0, 3fr);
   }
 
   .toc-desktop {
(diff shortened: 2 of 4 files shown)


_components/TableOfContents.css
```diff
@@ -9,7 +9,7 @@
 @media (min-width: 1024px) {
   .content {
     display: grid;
-    grid-template-columns: minmax(auto, 960px) 1fr;
```
josh-collinsworth, line 12: This bit was the problematic code; `auto` allows the element's ancestors to dictate width. Setting `0` as the minimum overrides this and forces the box to just take up the column given. (8:3 felt like the right ratio to me, but it could be a different ratio.)

_components/TableOfContents.css
```diff
@@ -51,6 +50,12 @@
   }
 }
 
+@media (min-width: 86rem) {
+  .content {
+    grid-template-columns: minmax(0, 1fr) minmax(20rem, max-content);
```
josh-collinsworth, line 55: This just lets the sidebar get wide on larger screens.

_includes/doc.tsx
```diff
@@ -20,7 +20,7 @@ export default function Doc(data: Lume.Data, helpers: Lume.Helpers) {
 
   return (
     <div id="content" className="content">
-      <div class="px-4 sm:px-5 md:px-6 max-w-full">
```
josh-collinsworth, line 23: Makes the main content go to the middle above ~1200px or so at default font size.

_components/Navigation.css
```diff
@@ -1,6 +1,6 @@
 .nav {
   height: 100%;
-  min-width: 280px;
+  min-width: 16rem;
```
josh-collinsworth, line 3: I think it's better to set width/height in `rem` wherever possible because if a user prefers a large font size, we should adjust the spacing to accommodate.

_includes/doc.tsx
```diff
@@ -20,7 +20,7 @@ export default function Doc(data: Lume.Data, helpers: Lume.Helpers) {
...
+      <div class="px-4 sm:px-5 md:px-6 w-full max-w-[60rem] mx-auto">
```
josh-collinsworth, line 23: Centers the main content when the screen is wide enough. If there's a reason we weren't doing this before, we can change this.

thisisjofrank (approved):
Noice, good catch!
OOI, how come I couldn't see it happening?

PR #1463, opened 2025-02-20
move file

thisisjofrank:
move example file to scripts and add it to examples list for examples landing page
also update node import example


marvinhagemeister (approved):
gotta move it, move it 🎶

PR #1466, opened 2025-02-21
docs: add selector tips to lint plugins

marvinhagemeister:
- Point to `typescript-eslint`'s playground when developing plugins
- Add note about `:exit` pseudo

### DIFF
--- a/runtime/reference/lint_plugins.md
+++ b/runtime/reference/lint_plugins.md
@@ -142,7 +142,7 @@ tree instead of when going down.
 We highly recommend using the
 [typescript-eslint playground](https://typescript-eslint.io/play/) when
 developing lint rules. It allows you to inspect code and the resulting AST
-format. This makes it easy to see which selectors match which node a lot easier.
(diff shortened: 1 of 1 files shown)


runtime/reference/lint_plugins.md
```diff
@@ -133,6 +133,19 @@ supported syntax for selectors is:
...
+There is also the `:exit` pseudo that is only valid at the end of the whole
+selector. When it's present, Deno will call the function while going **up** the
+tree instead of when going down.
+
+:::tip
+
+We highly recommend using the
+[typescript-eslint playground](https://typescript-eslint.io/play/) when
+developing lint rules. It allows you to inspect code and the resulting AST
+format. This makes it easy to see which selectors match which node a lot easier.
```
thisisjofrank, line 145:
```suggestion
format. This makes it easier to see which selectors match which node.
```

thisisjofrank (approved):
Looks good after that one change

PR #1469, opened 2025-02-21
Joco/update lts

josh-collinsworth:


runtime/fundamentals/stability_and_releases.md
```diff
@@ -40,9 +40,9 @@ compatible bug fixes.
 
 ![Deno long term support schedule](./images/deno-lts-support.png)
 
-A new LTS version is twice per year and **supported for 6 months**, giving users
-a month long window to upgrade between LTS relases. A new patch version of an
-LTS channel will be released at least **once per quarter**.
+A new LTS version is released every year, and **supported for a full year**,
+giving users a three-month window to upgrade between LTS releases. A new patch
+version of an LTS channel will be released at least **once per quarter**.
```
ry, line 45: I don't want to change the timeline here, only the minor releases are every 12 weeks instead of every 6 weeks.

josh-collinsworth: So the LTS version still only lasts six months, then?

ry: Correct

ry (approved):
LGTM

PR #1471, opened 2025-02-21
fix: remove reference to jsr:@db/sqlite

dannylindquist:
I might've been wrong, but this comment seemed stale now.


marvinhagemeister (approved):
Good spot, thanks!

PR #1480, opened 2025-02-23
Added documentation how to set up deno lsp in kickstart.nvim

nikkez:
Documentation added how to set up deno lsp when using the [kickstart.nvim](https://github.com/nvim-lua/kickstart.nvim) neovim configuration template and Mason LSP.


marvinhagemeister (approved):
Thanks!

evanlesmez:
So for ts and deno non-conflicting support, configuration would then look like this
```lua
         ts_ls = {
            workspace_required = true,
            root_markers = { 'package.json' },
          },
          denols = {
            workspace_required = true,
            root_markers = { 'deno.json', 'deno.jsonc' },
          }
```

PR #1482, opened 2025-02-24
hide uncategorized and make deno namespace the langing page for api/deno

thisisjofrank:
hide uncategorized category under the /api/deno section 
make deno namespace the langing page for api/deno
make width of sidebar not go below 250px so that it is the same across sections.


crowlKats (approved):
LGTM!

PR #1488, opened 2025-02-25
make examples pages wide

thisisjofrank:
Old:
image

New:
image


philhawksworth (approved):
Nice.

Ideally we'd make this a CSS variable at some point for easier consistency. But that's not a blocker to ship.

PR #1489, opened 2025-02-25
Add simple intro paragraph

josh-collinsworth:
Also moves the Node built-ins link further down, to a more relevant section.

CleanShot 2025-02-25 at 11 59 14@2x

### DIFF
--- a/runtime/fundamentals/node.md
+++ b/runtime/fundamentals/node.md
@@ -23,7 +23,7 @@ oldUrl:
 
 For example, here's how you'd import Hono from npm in a Deno project:
 
-```
+```ts
 import { Hono } from "npm:hono";
 ```
 
--- a/runtime/fundamentals/node.md
+++ b/runtime/fundamentals/node.md
@@ -18,8 +18,8 @@ oldUrl:
 
 - **Deno is Node-compatible**. Most Node projects will run in Deno with little
   or no change!
-- **Deno supports npm packages**. Just use the `npm:` specifier in
-  the import, and Deno takes care of the rest.
+- **Deno supports npm packages**. Just use the `npm:` specifier in the import,
+  and Deno takes care of the rest.
 
 For example, here's how you'd import Hono from npm in a Deno project:
 
--- a/runtime/fundamentals/node.md
+++ b/runtime/fundamentals/node.md
@@ -16,12 +16,20 @@ oldUrl:
 - /runtime/manual/node/private_registries
 ---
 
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/node.md
```diff
@@ -16,12 +16,20 @@ oldUrl:
 - /runtime/manual/node/private_registries
 ---
 
-Modern Node.js projects will run in Deno with little to no reworking required.
-However, there are some key differences between the two runtimes that you can
-take advantage of to make your code simpler and smaller when migrating your
-Node.js projects to Deno.
+- **Deno is fully Node-compatible**. Virtually any modern Node project will run
+  in Deno with little or no change!
```
philhawksworth, line 20: Having "fully" and then "virtually" together seems like a slight contradiction.
I think we can simplify this a touch and sound less like we are equivocating. How about this, which removes the "virtually" caveat and refocusses on how most project require little or no change?

```suggestion
- **Deno is Node-compatible**. Most Node projects will run in Deno with little
  or no change!
```

runtime/fundamentals/node.md
```diff
@@ -16,12 +16,20 @@ oldUrl:
...
+- **Deno supports virtually all npm packages**. Just use the `npm:` specifier in
+  the import, and Deno takes care of the rest.
```
philhawksworth, line 22: I'd also remove the"virtually" caveat here from this headline. We can expand into talking about some limitations in the more detailed part of the page, but the headline here is "you can use npm packages in deno, and look at the simple syntax".

How about:

```suggestion
- **Deno supports npm packages**. Just use the `npm:` specifier in
  the import, and Deno takes care of the rest.
```

runtime/fundamentals/node.md
```diff
@@ -16,12 +16,20 @@ oldUrl:
...
-Node.js projects to Deno.
+- **Deno is Node-compatible**. Most Node projects will run in Deno with little
+  or no change!
+- **Deno supports npm packages**. Just use the `npm:` specifier in the import,
+  and Deno takes care of the rest.
 
-<a href="/api/node/" class="docs-cta runtime-cta">Explore built-in Node APIs</a>
+For example, here's how you'd import Hono from npm in a Deno project:
+
+```
```
philhawksworth, line 26: Not sure if the dark mode version of this has the bold border due to the lack of filetype being specified in the code block... but we could add it anyway 

```suggestion
```ts
```

philhawksworth (approved):
LGTM

PR #1491, opened 2025-02-25
add trailing slashes to make sure urls match for active styling

thisisjofrank:


_components/SecondaryNav.tsx
```diff
@@ -92,7 +93,7 @@ export default function (data: Lume.Data) {
                     <a
                       href={item.href}
                       className="sub-nav-link blocklink"
-                      data-active={item.href === currentUrl}
+                      data-active={item.href.replace(/\/$/, "") === currentUrl}
```
bartlomieju, line 96: IMHO it would be better to go through all these items at the top of the function and check that each `href` ends in a trailing slash - otherwise throw an error - that way we can catch "bad hrefs" during development without doing computation on each render (that could potentially get expensive at some point). WDYT?

philhawksworth: Just making sure that I understand your concerns regarding "each render".

This component is used during the site generation in the build step and not client-side so abstracting out a data check would need to come out of the component completely. I think that would introduce more logical friction than we'd gain from a speed boost for site generation time.

PR #1497, opened 2025-02-26
fix: raise nav sidebar z-index

nagarete:
Prior to this commit, when the viewport is less than 1024px, the titles of some snippets would go on top of the sidebar. This raises the z-index of the nav sidebar so that it will always go over these texts.

### nav closed
before

### nav opened
after

### DIFF
--- a/_components/Navigation.css
+++ b/_components/Navigation.css
@@ -17,6 +17,7 @@
     transition: transform 200ms ease-in-out;
     transform: translateX(-100%);
     background-color: hsl(var(--background-primary));
+    z-index: 11;
   }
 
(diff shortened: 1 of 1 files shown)


_components/Navigation.css
```diff
@@ -17,6 +17,7 @@
     transition: transform 200ms ease-in-out;
     transform: translateX(-100%);
     background-color: hsl(var(--background-primary));
+    z-index: 11;
```
nagarete, line 20: The problematic snippet titles have a `z-10` tailwind class. I figured I just put 1 number greater.

PR #1498, opened 2025-02-26
fix: change websockets url

nagarete:
Prior to this commit, the link to the websockets API page was broken (404). It was supposed to be `https://docs.deno.com/api/deno/websockets` instead of `https://docs.deno.com/api/deno/web-sockets`.


thisisjofrank (approved):

thisisjofrank:
Nice catch!

PR #1505, opened 2025-02-27
Fix centered layout and tweak other small things

josh-collinsworth:
Also makes markdown filenames a little more readable (before the background was transparent, and the alighnment was a little off).

### DIFF
--- a/_components/Footer.css
+++ b/_components/Footer.css
@@ -46,7 +46,7 @@ footer {
   }
 
   .footer-nav {
-    max-width: 1200px;
+    max-width: 75rem;
     margin: 0 auto;
   }
 }
--- a/_components/TableOfContents.css
+++ b/_components/TableOfContents.css
@@ -12,10 +12,6 @@
     grid-template-columns: minmax(0, 8fr) minmax(0, 3fr);
   }
 
-  .examples-content {
-    max-width: 1200px;
-  }
-
   .toc-desktop {
     display: block;
     position: sticky;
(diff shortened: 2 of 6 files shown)


philhawksworth:
I think we need to try for a more consistent feel to the max content width across the different content types and variants with or without a ToC.

It feels odd to me that some of these appear to have quite narrow content even though the space is available to make them consistent with other pages:

Screenshot 2025-02-28 at 11 47 09

Screenshot 2025-02-28 at 11 47 04

Screenshot 2025-02-28 at 11 46 56

josh-collinsworth:
> It feels odd to me that some of these appear to have quite narrow content even though the space is available to make them consistent with other pages:

So the maximum width I set on narrower screens is based on what that content is _already_ allowed to stretch to. Paragraphs have a maximum width of 75rem by default, and wrap there no matter how wide the content is. So I just capped non-example pages at that same maximum width, since it would look weird to go wider but still have paragraphs wrap "early."

I think we _want_ those examples to be wider, because we have two columns of info. But when we only have one column, I don't think it makes sense to take up more space just for the sake of making pages uniform.

tl;dr: absent any ideas for how to make the max width uniform without unnecessarily stretching some kinds of content or truncating others, I think the inconsistency across pages is better than the alternative.

josh-collinsworth:
@philhawksworth @thisisjofrank Any additional thoughts on this? I personally think a layout of inconsistent width is better than one with mixed fullwidth and left-cropped items, but I'm open to other suggestions.

thisisjofrank (approved):
Yeah, looks good to me!

PR #1507, opened 2025-02-28
Tanstack: Add brackets for origin field when setting CORS

hreynier:
## What?

For the tanstack tutorial, add brackets for the origin field when setting cors in the Hono api.

## Why?

Otherwise, you get cors errors.
There's a [forum post](https://questions.deno.com/m/1311051310673690765) that encountered the issue, but setting the origin as an array fixes it.

### DIFF
--- a/examples/tutorials/tanstack.md
+++ b/examples/tutorials/tanstack.md
@@ -83,7 +83,7 @@ const app = new Hono();
 app.use(
   "/api/*",
   cors({
-    origin: "http://localhost:5173",
+    origin: ["http://localhost:5173"],
     allowMethods: ["GET", "POST", "PUT", "DELETE"],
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
The `origin` value of Hono's `cors` middleware can be a `string`, `string[]` or a `function`. See https://github.com/honojs/hono/blob/87963713c5db1a86a460195ff131fa82e5af9a63/src/middleware/cors/index.ts#L10

```ts
type CORSOptions = {
  origin: string | string[] | ((origin: string, c: Context) => string | undefined | null)
  // ...
}
```

If you're getting CORS errors then it's related to something else. Whether you pass the origin as a `string` or as a `string[]` makes no difference from Hono's perspective.

Issue #1265, opened 2024-12-15
Garbled docs for deno run --reload


dandv:
At https://docs.deno.com/runtime/reference/cli/run/#dependency-management-options-reload

![image](https://github.com/user-attachments/assets/c0fbd90b-ce12-497a-877d-bec55ccdb38c)

I tried to fix it myself but the "Edit this page" link at the bottom points to https://github.com/denoland/docs/edit/main/runtime/reference/cli/run.md, which looks like a completely different file.

thisisjofrank:
These pages are generated from the CLI --help output
Looks like the tabular data from the output isn't being formatted correctly, hence the non-sensical reading order.

![image](https://github.com/user-attachments/assets/de9b051e-1270-4087-98f6-be7b6514be34)

dandv:
> These pages are generated from the CLI --help output

Would it be possible to add a note to the CLI .md files explaining this, so contributors know where to submit contributions?

PR #1510, opened 2025-03-02
docs: fix punctuation in _commands_reference.json

dandv:
Hopefully that will be reflected at https://docs.deno.com/runtime/reference/cli/lint/#options-config.

Could a comment header please be added to https://github.com/denoland/docs/edit/main/runtime/reference/cli/lint.md (and similar files that silently transclude Markdown from other sources) explaining how the entire page is generated?

I keep running into this, and being a drive-by docs contributor, I keep forgetting:
https://github.com/denoland/docs/issues/1265#issuecomment-2547277394

### DIFF
--- a/runtime/reference/cli/_commands_reference.json
+++ b/runtime/reference/cli/_commands_reference.json
@@ -184,7 +184,7 @@
           "short": "c",
           "long": "config",
           "required": false,
-          "help": "Configure different aspects of deno including TypeScript, linting, and code formatting\n  \u001b[38;5;245mTypically the configuration file will be called `deno.json` or `deno.jsonc` and\n  automatically detected; in that case this flag is not necessary.\n  Docs: https://docs.deno.com/go/config\u001b[39m",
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
FYI: This file is autogenerated by https://github.com/denoland/docs/blob/d999e31158ac31764d9c19fa583cd4368e51f54e/.github/workflows/update_versions.ts#L25-L27

It's basically the output of `deno -h`. Any changes to this file here will be lost by the next Deno update. The PR should be made against this line in the Deno repo where the message is coming from:

https://github.com/denoland/deno/blob/c2abcb9a3de77d3c8ee502d12ec5a27c55e17c9c/cli/args/flags.rs#L4204-L4207

thisisjofrank:
https://github.com/denoland/deno/pull/28447 updated the cli output

dandv:
@thisisjofrank thanks for updating the source. How would one go about adding comments to these autogenerated .md files informing contributors that they're autogenerated, and where to contribute?

thisisjofrank:
the problem is that they're built from d.ts files which are spread throughout the deno codebase. So it isn't possible to go to one place to look for them, even using grep will often return multiple instances of a string of text in the Deno repo as some types have the same description or are used in multiple places.
As the contribution page says "If you find an issue with one of these pages, you can either submit a pull request to the Deno repository, or raise an issue on the docs site and we'll fix it." That is kind of the best advice we can give.

Issue #1513, opened 2025-03-02
Filtering the list of linting rules doesn't reset when deleting the last character in the filter


dandv:
1. Go to https://docs.deno.com/lint/
2. Type `z` in the "Search lint rules" input field.
3. Delete the `z`

![Image](https://github.com/user-attachments/assets/dde29081-e31b-4623-b960-ea6fb7afc13b)

PR #1519, opened 2025-03-03
fix: lint rule search not cleared on empty input

marvinhagemeister:
The search value wasn't reset internally when it had a value and was cleared again.

Fixes https://github.com/denoland/docs/issues/1513


philhawksworth (approved):
LGTM!

PR #1520, opened 2025-03-03
fix: lint rule items jumping around on search

marvinhagemeister:
The lint rule filtering page inherits styles from `gfm.css`. That file includes a rule like this:

```css
.markdown-body li + li {
  margin-top: ...
}
```

Which gives every lint rule item except the first a top margin. When filtering the rule list the first item is often hidden, which means that the first _visible_ item got a margin top and moved the result list around.


philhawksworth (approved):
👍

PR #1522, opened 2025-03-03
code: updated fetch call after experiencing "Failed to fetch dinosaurs"

luhouyang:
The extra slash at the end of fetch call caused request to resolve incorrectly, resulting in **"Failed to fetch dinosaurs"**.

Changes in: `./src/components/DinosaurList.tsx (line 5)`

Original code
---------------
const response = await fetch("http://localhost:8000/api/dinosaurs/");

Proposed change
-------------------
const response = await fetch("http://localhost:8000/api/dinosaurs");


thisisjofrank (approved):
Good catch!

PR #1523, opened 2025-03-04
docs: fix incorrect importMap description

fkywong:
Per discussion in denoland/deno#28380, the actual behavior of `importMap` conforms to the Import Maps Standard, rather than the Deno extension that does away with these double entries.


dsherret (approved):
Thanks!

PR #1524, opened 2025-03-04
fix: correct code block from reference docs

tugrulates:
There is a bug in calculating whether to use inline or pre code blocks.

Example: https://docs.deno.com/runtime/reference/cli/install/
image

The offset is used to determine where the ANSI color block match is in the reference string. If the line is a full match, a pre code block is rendered, otherwise an inline code block is rendered.

Line 87 loses the linefeed characters used in the split. The calculation was deviating slightly from the correct offset as a result. In the edge case for the inline code block near the end of a line, its type is derived from the next line, and incorrectly when the next line is a pre block.

Fix:
image

Tested by building locally, and spot checking other commands.

### DIFF
--- a/_includes/renderCommand.tsx
+++ b/_includes/renderCommand.tsx
@@ -89,7 +89,7 @@ export default function renderCommand(
 
         while (offset > 0) {
           line = lines.shift();
-          offset -= line.length;
+          offset -= line.length + 1;
         }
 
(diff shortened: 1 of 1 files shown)


thisisjofrank (approved):

thisisjofrank:
This is a great catch, thank you! That's been on my tofix list for a while! What a nice solution

PR #1527, opened 2025-03-04
Generate OG images for each page

philhawksworth:
- Adds `/styleguide/og/` page for testing and developing OG image layouts
- Generates a png for each URL in the site taking in the title and description

Todo:

- [x] Design review and tuning
- [x] Handle empty descriptions
- [x] Associate with the page template


josh-collinsworth:
Kind of a bummer there's no way to get the generator to recognize our font settings, but it doesn't look like the under-the-hood tool has any support for advance font features, so not much we can do there (short of cutting our own font file with those features as the default, anyway).

CleanShot 2025-03-06 at 16 10 36@2x

josh-collinsworth:
Hope these changes are welcome! I'm not entirely sold on the border on the image, but I think it looks better in light mode, at least. I went back and forth on the gradient background, but ultimately I think it just generally looks better than the solid Deno color. I could maybe be persuaded a solid color is better if we went with a lighter one.

At any rate, unless there are any major objections, I think this is ready to go, and we can iterate if needed.

thisisjofrank (approved):
Looks good to me!

PR #1536, opened 2025-03-06
Additional example and description for setting en vars when running a script

philhawksworth:
Address docs feedback:

> how to run deno tasks with envs set, it doesnt seem to appear when I call Deno.env.get(ENV_NAME)  

Adds more descption and an example with `deno task`

### DIFF
--- a/runtime/reference/env_variables.md
+++ b/runtime/reference/env_variables.md
@@ -80,6 +80,9 @@ variable, and can be helpfully combined with
 
 ```json title="deno.json"
 {
+
+  ...
+  
   "tasks": {
     "build:full": {
       "description": "Build the site with all features",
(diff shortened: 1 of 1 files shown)


runtime/reference/env_variables.md
```diff
@@ -65,6 +65,37 @@ console.log(Deno.env.get("GREETING")); // "Hello, world."
...
+
+```shell
+MY_VAR="my value" deno run main.ts
+```
+
+This can be useful when you want to vary a task based on an environment
+variable, and can be helpfully combined with
+[`deno task`](/runtime/reference/cli/task/) commands like so:
+
+```json title="deno.json"
```
thisisjofrank, line 81:
```suggestion
```jsonc title="deno.json"
```

PR #1543, opened 2025-03-10
docs(completions): support other Linux distros

dandv:
On Fedora, bash completions live at `/usr/share/bash-completion/completions/`.


thisisjofrank (approved):
LGTM

PR #1550, opened 2025-03-12
updating config docs to make it clearer that there are other fields a…

thisisjofrank:
…vailable, and fixing some formatting


philhawksworth (approved):
LGTM

PR #1553, opened 2025-03-12
show how to include or exclude any lint rule in deno.json

philhawksworth:
1. Adds information to every lint rule page to show how to use the `include` and `exclude` arrays in deno.json in order to configure if this rule is used on not in `deno lint`

For example on `/lint/rules/adjacent-overload-signatures`

Screenshot 2025-03-12 at 15 26 17

2. States what the default set of lint tags are if none is provided
3. Adds links to more description about the difference between `lint` and `fmt` in the pages for both cli commands
4. Provides more obvious link from the rules list to the `deno.json` config that allows inclusion and exclusion


thisisjofrank (approved):
This looks good to me

PR #1556, opened 2025-03-13
Homepage components

thisisjofrank:


index.mdx
```diff
@@ -0,0 +1,143 @@
+---
+title: "Deno Documentation"
+description: "The best docs ever"
```
philhawksworth, line 3: I like this aspiration, but should we be more descriptive here. Especially now that it surfaces not only for search engines but also in the OG image when people share a link to our docs homepage

https://deno-docs--homepage-components.deno.dev/index.png

styleguide/components.mdx
```diff
@@ -35,6 +35,52 @@ Decorator for external links
   deno.com
 </comp.ExternalLink>
 
+## Hero
+
+A hero section for the top of a page
+
+```tsx
+<comp.Hero  bgImage="deno-looking-up.svg">
```
philhawksworth, line 43: double space here, but also this image doesn't appear in the example. Does it need a fix for the path to the image?

```suggestion
<comp.Hero bgImage="deno-looking-up.svg">
```

philhawksworth (approved):
YES! The componentisation continues 🥳
A couple of small, non-blocking comments from me.

index.mdx
thisisjofrank: Oop yeah! That's what I get for not using lorem ipsum placeholders!

PR #1568, opened 2025-03-17
adding example for createObjectURL

thisisjofrank:
Makes a cute little demo 
image

### DIFF
--- a/examples/scripts/create_object_url.ts
+++ b/examples/scripts/create_object_url.ts
@@ -1,8 +1,8 @@
 /**
- * @title Create Object URL
+ * @title Generate downloadable files
  * @difficulty beginner
  * @tags web
- * @run -allow-net
+ * @run --allow-net
  * @resource {https://docs.deno.com/api/web/~/URL.createObjectURL } Doc: Deno.createObjectURL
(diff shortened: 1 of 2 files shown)


examples/scripts/create_object_url.ts
```diff
@@ -0,0 +1,185 @@
+/**
+ * @title Create Object URL
```
ry, line 2: createObjectURL

examples/scripts/create_object_url.ts
```diff
@@ -0,0 +1,185 @@
...
+ * @difficulty beginner
+ * @tags web
+ * @run -allow-net
```
ry, line 5: `--allow-net`

ry (commented):
I think an example should show usage in deno, not client side

This is a pretty niche API mostly here for compatibility reasons, I'm not sure it warrants an example. The one place I've seen it is when creating a Web Worker and wanting to define the worker's source code in the same file, you can use URL.createObjectURL to pass to the worker constructor - [example here](https://github.com/denoland/deno/issues/9210)
I'd just drop that example into the JSDoc and probably leave out the full example page, since its uncommonly used.

PR #1577, opened 2025-03-20
make og image url absolute

thisisjofrank:


satyarohith (approved):
RSLGTM

if you're expecting og:url tag to be set then it's not set in both [docs.deno.com](https://orcascan.com/tools/open-graph-validator?url=docs.deno.com) and the [preview domain](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fdeno-docs-d19dg46x5h66.deno.dev%2F)

PR #1582, opened 2025-03-21
fix: deno serve example with json conditional

fabricionaweb:
`request.url` is returning a full url: `http://localhost:3000/json`
`startsWith` wont match, but `endsWith` does


marvinhagemeister (approved):
Good catch, thanks!

PR #1583, opened 2025-03-22
Update Temporal Docs

Mrashes:
This updates the example to console log values to show that its actually working as expected.

In addition there is a problem where the run prompt doesn't work without the flag but this will at least show that the script worked when the user passes the flag.


Mrashes:
Sorry about the bad build - edited on github so didn't have linting/formatting built in.

PR #1598, opened 2025-03-31
docs: fix docker healthcheck example

TorstenDittmann:
Before:

```sh
deno eval "try { await fetch('http://localhost:8000/health'); } catch { exit(1); }"
error: Uncaught (in promise) ReferenceError: exit is not defined
try { await fetch('http://localhost:8000/health'); } catch { exit(1); }
                                                             ^
    at file:///Users/torstendittmann/Documents/GitHub/TorstenDittmann/contained/$deno$eval.mts:1:62
    at eventLoopTick (ext:core/01_core.js:178:7)
```

Using `Deno.exit(1)` will make this work 👍🏻

### DIFF
--- a/runtime/reference/docker.md
+++ b/runtime/reference/docker.md
@@ -151,7 +151,7 @@ services:
 
 ```dockerfile
 HEALTHCHECK --interval=30s --timeout=3s \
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good catch, thanks for making a PR!

PR #1603, opened 2025-04-01
fix: replace reference to Mimir with Prometheus

rauno56:
The description of LGTM stack mentions Mimir which is in fact a component in Grafana stack, but probably not the metrics component that was meant.

### DIFF
--- a/examples/_data.ts
+++ b/examples/_data.ts
@@ -325,6 +325,11 @@ export const sidebar = [
         externalURL: "https://www.youtube.com/watch?v=P2BBYNPpgW8",
         type: "video",
       },
+      {
+        title: "Testing web applications",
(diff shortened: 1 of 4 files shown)


lucacasonato (commented):
I think the LGTM stack does actually use Mimir: https://grafana.com/go/webinar/getting-started-with-grafana-lgtm-stack/

Maybe change it to `Mimir (metrics, Prometheus compatible)`?

wperron:
While Grafana does use Mimir in production in Grafana Cloud, it looks like their all-in-one LGTM Docker image [actually uses Prometheus](https://github.com/grafana/docker-otel-lgtm/blob/main/docker/Dockerfile#L34) so LGTM

PR #1625, opened 2025-04-16
docs: add caution callout for kv

cddm-gh:
Hi, I was trying out the `Deno.openKv()` feature today, and I was getting this error:
`TypeError: Deno.openKv is not a function.` After investigating, I realized I was missing a flag mentioned [here](https://github.com/denoland/deno/issues/25638#issuecomment-2351012531), so I think it's a good idea to let people know when they're reading the documentation for this feature and avoid wasting time afterward.

This is how it's going to look:
image


deploy/kv/manual/index.md
```diff
@@ -6,6 +6,12 @@ oldUrl:
   - /runtime/manual/runtime/kv/
 ---
 
+:::caution This feature is still unstable
+
+Make sure you add the `--unstable-kv` flag to your run command to enable it.
+
+:::
```
marvinhagemeister, line 13: Nit: let's use a bit softer language:
```suggestion
:::caution

Deno KV is still in development and may change. To use it, you must pass the `--unstable-kv` flag to Deno.

:::
```

thisisjofrank (approved):
Thanks for spotting this, looks like the admonition got emptied here!

PR #1636, opened 2025-04-17
Various small fixes

philhawksworth:
A few small updates based on some user feedback / questions

- path operation examples
- mentioning deprecation of deno cache
- lcov explanation in coverage
- mention `npx` and how we do the equivalent in out Node compatibility page

### DIFF
--- a/runtime/fundamentals/node.md
+++ b/runtime/fundamentals/node.md
@@ -97,6 +97,17 @@ npm specifiers have the following format:
 npm:<package-name>[@<version-requirement>][/<sub-path>]
 ```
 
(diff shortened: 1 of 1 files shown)


thisisjofrank (approved):
Lgtm. Might consider changing the order of the dependency management list eventually to move cache lower.

philhawksworth:
> Lgtm. Might consider changing the order of the dependency management list eventually to move cache lower.

Ta!

Yeah, I stuck with the alphabetical order for now. Perhaps if we end up with more deprecated items there in the future, we might think about grouping those differently to keep things clean.

PR #1640, opened 2025-04-17
Align typeparam description to TSDoc

tymokvo:
## Summary
- Align description of `typeparam` tag to TSDoc

## Description
I noticed today while browsing the ["Supported Tags" section of `deno doc`](https://docs.deno.com/runtime/reference/cli/doc/#supported-tags) docs that "define a callback" seems to have been copypasta'd into the `typeparam` tag.

I updated this to match the description from the linked TSDoc page.


thisisjofrank (approved):
Good spot!

PR #1641, opened 2025-04-17
test revert trialing slash

philhawksworth:
Issue seems to have arrive at #1526 

All works locally and page created ok, so testing if the URL is treated differently in prod vs local as this appears to be the only things in that change that might have an impact.


philhawksworth:
Nope.

hmmm... The page is created correctly:
https://deno-docs-khtgqhjz2e7t.deno.dev/404

PR #1644, opened 2025-04-19
docs: Fix cron schedule backoff example

NuroDev:
Currently the `Deno.cron` documentation for retrying failed runs ([Source](https://docs.deno.com/deploy/kv/manual/cron/#retrying-failed-runs)) shows a code snippet that is, I am assuming, out of date as it recommends setting the options object as the fourth parameter when in reality it should be the third parameter & the last parameter is always set to be the cron handler.


NuroDev:
After a bit more research, yes. It does seem this function overload was recently removed in denoland/deno#22035.

thisisjofrank (approved):
Good spot, thank you!

PR #1646, opened 2025-04-21
Update import to use package on jsr.io

IngoHohmann:
Update duckdb example to use package on jsr.io


marvinhagemeister (approved):
Thanks!

PR #1651, opened 2025-04-23
update vite tut to add type support

thisisjofrank:


philhawksworth (approved):
Great! 
This all reads well and everything runs and behaves as expected for me when I work through the example

PR #1654, opened 2025-04-23
adding a little more information to the workspace docs

thisisjofrank:


runtime/fundamentals/workspaces.md
```diff
@@ -383,3 +383,137 @@ root and its members:
...
+    "build": "echo 'Building add package'",
+    "test": "deno test"
+  }
+}
+```
+
+To run a task defined in a specific package:
+
+```sh
+deno task --cwd=add build
```
dsherret, line 463: I'm actually surprised this affects which task is resolved. I tried this out and it does work though.

runtime/fundamentals/workspaces.md
```diff
@@ -383,3 +383,137 @@ root and its members:
...
+
+This makes these dependencies available to all workspace members without needing
+to redefine them.
+
+### Managing Version Conflicts
+
+The workspace root's import map takes precedence over member-specific imports,
+which helps ensure consistent versions across your workspace. For example, if
+both the root and a member specify different versions of the same dependency,
+the root's version will be used.
```
dsherret, line 494: It's the opposite. It will use the member's version when resolving the dependency within the member's folder. The member allows overwriting it. That said, when resolving dependency's outside the member's folder, it will resolve based on the workspace root's folder.

Member deno.json dependencies are restricted to the specific member's folder, but the workspace root's deno.json applies to the entire application (including jsr and https dependencies)

dsherret (approved):
LGTM! But there is one issue under "Managing Version Conflicts"

runtime/fundamentals/workspaces.md
thisisjofrank: Have updated to:

When resolving dependencies, workspace members can override dependencies defined
in the root. If both the root and a member specify different versions of the
same dependency, the member's version will be used when resolving within that
member's folder. This allows individual packages to use specific dependency
versions when needed.

However, member-specific dependencies are scoped only to that member's folder.
Outside of member folders, or when working with files at the workspace root
level, the workspace root's import map will be used for resolving dependencies
(including JSR and HTTPS dependencies).

PR #1664, opened 2025-04-25
update permissions samples on testing.md

kuboon:
I was confused by the permissions option in Deno.test, so I’d like to suggest a correction.

In the early example, permissions: { read: true }, is included, but the test actually passes even without it. I found this point unclear, so I think it should be removed.

A more detailed explanation is provided in the ## Tests and Permissions section, but including the unnecessary permissions: { read: true }, there as well makes the line "allows you to specifically deny permissions, but does not grant them." harder to understand.

What do you think?

### DIFF
--- a/runtime/fundamentals/testing.md
+++ b/runtime/fundamentals/testing.md
@@ -48,7 +48,6 @@ Deno.test("async test", async () => {
 
 Deno.test({
   name: "read file test",
-  permissions: { read: true },
   fn: () => {
     const data = Deno.readTextFileSync("./somefile.txt");
     assertEquals(data, "expected content");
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/testing.md
```diff
@@ -97,7 +96,7 @@ deno test my_test.ts -- -e --foo --bar
 
 # Provide permission for deno to read from the filesystem, which is necessary
 # for the final test above to pass
-deno test --allow-read my_test.ts
+deno test --allow-read=. my_test.ts
```
kuboon, line 99: This is a bit of a personal preference, but I have a strong desire to promote the practice of not granting broad permissions too easily.

thisisjofrank (approved):
LGTM, thank you!

Issue #222, opened 2023-12-07
Provide rationale for prohibiting arrow syntax on top-level


scarf005:
https://github.com/denoland/deno-docs/blob/2c90cc2eca70c484e5fbfc61d256bf7342a2e95e/runtime/manual/references/contributing/style_guide.md?plain=1#L330-L349

While other rules also describe rationale for the rule, this section doesn't. Adding one could help understanding the rule better.

PR #1685, opened 2025-05-06
Give a little explanation of why no top level arrow functions

thisisjofrank:
Closes https://github.com/denoland/docs/issues/222

### DIFF
--- a/runtime/contributing/style_guide.md
+++ b/runtime/contributing/style_guide.md
@@ -342,32 +342,6 @@ Deno.test("foo() returns bar object", function () {
 Note: See [tracking issue](https://github.com/denoland/deno_std/issues/3754) for
 more information.
 
-#### Top-level functions should not use arrow syntax
-
-Top-level functions should use the `function` keyword. Arrow syntax should be
-limited to closures.
-
-Bad:
-
-```ts
-export const foo = (): string => {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):

scarf005 (commented):
rather than why no top level arrow function, the section is removed; does this mean top level arrow function is allowed?

bartlomieju:
> rather than explaining why no top level arrow function, the section is removed; does this mean top level arrow function is allowed?

No, it's still preferred to use function declaration, but this section just induced confusion instead of being useful.

kt3k:
This document only defines the internal style used in Deno (The section is under contributing category). If we still prefer that style, shouldn't we still keep this section?

side note: google typescript style guide has a similar rule https://google.github.io/styleguide/tsguide.html#function-declarations

scarf005:
> If we still prefer that style, shouldn't we still keep this section?

if so, the section should also clearly explain why it's preferred over arrow function.

kt3k:
What kind of explanation do you expect to see?

I think this is just a preference of the team. I think a style guide can have such rule

dsherret:
In my view, the reason to use function declarations over variable declarations with an arrow function is clarity, consistency, and hoisting.

PR #1694, opened 2025-05-09
move mdx files back to md

thisisjofrank:
Unfortunately, the Title and ToC plugins we're using do not work with mdx.  This means that any pages that are mdx files do not get the table of contents or the headings with anchor tags.

 We will have to re-evaluate how we add more complicated components to the markdown. For now I have swapped the files back to .md files and just thrown some html in there to render the CTAs. 

I imagine the fix might be to make our own markdown-it plugins to render custom components.


philhawksworth (approved):
Oh darn!

Good catch, but this is a shame. We'll have to evaluate another approach to get the best of both worlds.

PR #1728, opened 2025-05-19
fix examples for deno fmt and Node.js fs imports

magurotuna:
This PR fixes several documentation issues:

1. Added the missing stdin indicator ("-") in the deno fmt pipe examples, which
   is required when passing content through stdin.
2. Corrected the Node.js fs import syntax in the interoperability example to use
   the proper default import pattern.
3. Fixed the escaped quotes example in the deno_fmt documentation to correctly
   show the initial string with escaped quotes before formatting.

Special thanks to Johan Dahl for bringing these issues to our attention via
email.


littledivy (approved):
LGTM

Issue #1688, opened 2025-05-07
Incoherent/contradictory LTS documentation


AntonOfTheWoods:
Version: Deno LTS

Following discussion with @bartlomieju and @philhawksworth on https://github.com/denoland/deno/issues/29108, it is still decidedly unclear what the actual policy on current and future LTS releases is.

The [official project documentation on the matter](https://docs.deno.com/runtime/fundamentals/stability_and_releases/) states the following:

```
Deno releases a new stable, minor version (eg. v2.1.0, v2.0.0) on a 12 week schedule.
...
Starting with Deno v2.1.0 (released in November 2024), Deno offers an LTS (long-term support) channel.
...
LTS releases occur every six months, with patch releases as needed for bug fixes.
```

Unfortunately, the official project page also has an image - updated 2025-05-06 - that is *presumably* also supposed to reflect official project policy:

![Image](https://github.com/user-attachments/assets/68577a60-fd33-4a02-9be3-f9fa33368af1)

There are several points to note about the image:
- 2.1 and 2.2 are both supposedly "LTS" and were not released 6 months apart.
- According to the image, 2.4 will be an LTS
- According to the image, 2.6, released 6 months after 2.4, will NOT be an LTS
- 2.7 *might* be an LTS - we don't know because (expected/projected) actual dates aren't written anywhere and the image doesn't go far enough to tell.

Other reasonably major platforms (e.g, Supabase) claim to want to base their (e.g, SB edge-runtime) releases on Deno LTS releases, so if we don't know when Deno has LTS releases, then we can't possibly know when they will release. They won't know when to plan work.

endoflife.date, which is what comes up first if I do a plain Google search for "Deno LTS" also has complete rubbish, so I'm not the only one confused. As an example, it claims 2.2 is now EOL.

# Suggestion

I suggest completely removing the image and replacing it with a plain text (markdown) table with actual dates according to the currently decided LTS policy. Many (most?) major projects do it this way. This should mean less work to actually update going forward and far less confusion. Obviously project policy changes but the current policy has never actually been clearly written on the official project page dedicated to explaining current LTS policy.

dsherret:
> 2.1 and 2.2 are both supposedly "LTS" and were not released 6 months apart.

The chart starts in February and doesn't include the months prior. Deno 2.1 was released in November and it has remained the LTS release for 6 months.

> According to the image, 2.6, released 6 months after 2.4, will NOT be an LTS
>
[...]

PR #1751, opened 2025-05-21
Swap LTS chart for a textual table

philhawksworth:
Conveying the details of the LTS maintenance periods via an image was not good for upkeep or for accessibility.
Replacing that image with a table for better clarity and convenience.

Visible at: `/runtime/fundamentals/stability_and_releases/`

closes:  #1688


marvinhagemeister (approved):
LGTM

PR #1753, opened 2025-05-21
llms.txt

donjo:
Adds llms.txt file generation to be served at docs.deno.com/llms.txt and docs.deno.com/llms-full.txt

towards: #1760


philhawksworth (approved):
Great. Thanks @donjo!

Let's ship this, although I wonder if we could lean more on Lume's mechanics to generate this for us rather than having a separate script to parse and write all this.

Noting for a future iteration that might make maintenance easier as the content and structure of the site evolves.

donjo:
@philhawksworth also of note, this doesn’t include the api docs part of the site so we may want to consider how/if that gets included as well.

PR #1754, opened 2025-05-22
Add docs for dev imports

lucacasonato:


runtime/fundamentals/modules.md
```diff
@@ -457,6 +457,69 @@ deno run --reload my_module.ts
...
+
+Sometimes dependencies are only needed during development, for example
+dependencies of test files or build tools. In Deno, the runtime does not require
+you to distinguish between development and production dependencies, as the
+[runtime will only load and install dependencies that are actually used in the
+code that is being executed](#why-does-deno-not-have-a-devimports-field).
+
+However, it can be useful to mark dev dependencies to aid people who are reading
+your package. When using `deno.json`, the convention is to add a `// dev`
+comment after any "dev only" dependency:
```
marvinhagemeister, line 470: Nit, this requires you to rename `deno.json` to `deno.jsonc`

marvinhagemeister: yeah but in Fresh we read `deno.json` to check if jsx is set correctly and stuff and we assume that `.json` means json and `.jsonc` means jsonc. There is probably more software out there with that assumption

runtime/fundamentals/modules.md
```diff
@@ -457,6 +457,69 @@ deno run --reload my_module.ts
...
+
+```json title="deno.json"
+{
+  "imports": {
+    "@std/fs": "jsr:@std/fs@1",
+    "@std/testing": "jsr:@std/testing@1" // dev
```
dsherret, line 476: Not a fan :P

dsherret (approved):
LGTM

runtime/fundamentals/modules.md
```diff
@@ -457,6 +457,69 @@ deno run --reload my_module.ts
 deno run --reload=jsr:@std/fs my_module.ts
 ```
 
+## Development only dependencies
+
+Sometimes dependencies are only needed during development, for example
+dependencies of test files or build tools. In Deno, the runtime does not require
+you to distinguish between development and production dependencies, as the
+[runtime will only load and install dependencies that are actually used in the
+code that is being executed](#why-does-deno-not-have-a-devimports-field).
```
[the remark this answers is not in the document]
lucacasonato: This is clarified in the last paragraph of "Why does Deno not have a `devImports` field?"

PR #1766, opened 2025-05-26
fix some mock tutorial examples

kt3k:
This PR fixes 2 examples in `examples/mocking_tutorial` page.

### DIFF
--- a/examples/tutorials/mocking.md
+++ b/examples/tutorials/mocking.md
@@ -56,7 +56,7 @@ function saveUser(
 // Test with a mock
 Deno.test("saveUser calls database.save", async () => {
   // Create a mock database with a spy on the save method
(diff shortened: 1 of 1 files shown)


examples/tutorials/mocking.md
```diff
@@ -351,11 +351,11 @@ class UserService {
   }
 }
 
-Deno.test("partial mocking with spies", async () => {
+Deno.test("partial mocking with stubs", async () => {
   const service = new UserService();
 
   // Only mock the getUser method
-  const getUserSpy = spy(
+  const getUserMock = stub(
```
kt3k, line 358: `spy` doesn't have capability of replacing the implementation. We need to use `stub` to replace it

Issue #1765, opened 2025-05-25
TCP Echo Server Netcat example doesn't appear work as expected


julillermo:
I'm a newbie programmer going through the [Deno examples](https://docs.deno.com/examples/).

I was attempting to recreate the ["TCP Echo Server" example](https://docs.deno.com/examples/tcp_echo_server/), and, I'm not sure, but it would seem that the use of Netcat isn't working as expected.

There is something about the exclamation point `!` in the provided `echo "Hello, Deno!" | nc localhost 8080` command seems to prevent it from working.

I'm not entirely sure as to why, but adding backslash `\` before `!` or removing the quotation marks for `Hello, Deno!` would return what I assume the expected result would be.

PR #1771, opened 2025-05-27
Use single quotes to avoid need to escape the ! in this example

philhawksworth:
Fixes confusion as reported in #1765

closes: #1765


thisisjofrank (approved):
LGTM

Issue #1773, opened 2025-05-28
bug: Deploy EA docs contains broken contact us link


iuioiua:
Under [Support and Feedback](https://docs.deno.com/deploy/early-access/#support-and-feedback), [contact us](https://docs.deno.com/deploy/support) links to a page that doesn't exist.

PR #1775, opened 2025-05-28
Fixed contact us link in Deploy Early Access Docs

andyvu923:
Towards #1773 

Quick fix to change link in [Support and Feedback](https://docs.deno.com/deploy/early-access/#support-and-feedback) so that link properly connects to support page.


lucacasonato:
Ah we already fixed this, thanks

PR #1784, opened 2025-05-29
Fix Astro init command

mintydev789:
The commands in the doc for creating an Astro project don't match. This PR fixes that in favor of the command that's closer to the one in the [Astro docs](https://docs.astro.build/en/install-and-setup/).


mintydev789:
@philhawksworth okay, thanks!

PR #1785, opened 2025-05-29
Add VS Code integration section to the formatter documentation

mintydev789:
I found it incredibly unintuitive how to enable Deno as my formatter in VS Code, and even opened [an issue](https://github.com/denoland/deno/issues/27999) about it before. It would be good to have this cleared up in the docs.


thisisjofrank (approved):
Thank you, good call

PR #1811, opened 2025-06-09
Deno Deploy EA updates for 9th of June 2025

lucacasonato:


deploy/early-access/reference/env-vars-and-contexts.md
```diff
@@ -43,6 +43,13 @@ Need more contexts? Please reach out to [support](../support).
 
 :::
 
+Additionally ther is also a **Build** context, which is used during the build
```
thisisjofrank, line 46:
```suggestion
Additionally, there is also a **Build** context, which is used during the build
```

PR #1814, opened 2025-06-10
Og html

thisisjofrank:
If there is html in the title or description, look for a sup, if its a sup convert the text contents to unicode , if not, strip out the html. 
This is ugly as hell, sorry.


philhawksworth (approved):
It'll do the job :)
No superscript character for `Q`?

PR #1818, opened 2025-06-11
Introduce `links` instead of `patch` in `deno.json` docs

temeddix:
With Deno 2.3.6, the `patch` field has been renamed to `links`. This PR updates the docs.

- https://github.com/denoland/deno/pull/29677

P.S. I would be grateful if the PR that makes `links` work with Vite get reviewed too.

- https://github.com/denoland/deno-vite-plugin/pull/69


thisisjofrank (approved):
Nice, thank you.

PR #1875, opened 2025-07-03
Update docker.md

kristianka:
Fixes Docker warning when building: 

```
1 warning found (use docker --debug to expand):  
  - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 2)
```


marvinhagemeister (approved):
Thanks!

PR #1888, opened 2025-07-08
adding a deploy button which links to the dashboard

thisisjofrank:


examples/tutorials/next.md
```diff
@@ -277,6 +277,24 @@ details!
 
 ![demo of the app](./images/how-to/next/dinoapp.gif)
 
+## Deploy your app
+
+Now that you have a Next.js app, you can deploy it to the edge with Deno
+Deploy<sup>EA</sup>. Create a [new GitHub repository](https://github.com/new),
+then initialize and push your app to GitHub:
```
philhawksworth, line 284: I suggest we don't say "deploy to the edge" since that is not always going to be the case. "deploy to the web" seems more tangible anyway,

I also think we should embellish this info just a little to indicate why the code needs to be published to GitHub first... perhaps:

```suggestion
Now that you have a Next.js app, you can deploy it to the web with Deno
Deploy<sup>EA</sup>. Create a [new GitHub repository](https://github.com/new),
then initialize and push your app to GitHub to be ready to use in a new application in Deno Deploy:
```

philhawksworth (approved):
LGTM!

PR #1898, opened 2025-07-10
update EA Deno version

ry:
this is going to age poorly. it's useful to have the version in the docs, but it's very hard to keep updated. almost seems worth adding some automation, but probably better to just remove this line?


thisisjofrank:
updated in another branch

PR #1909, opened 2025-07-15
Svelte demo

thisisjofrank:


examples/tutorials/svelte.md
```diff
@@ -0,0 +1,334 @@
...
+  if (dinosaur) {
+    return json(dinosaur);
+  }
+
+  return json({ error: "Not found" }, { status: 404 });
+};
+```
+
+SvelteKit automatically handles routing based on the file structure. The
+`+server.js` and `+server.ts` files define API endpoints, and the `[id]` folder
```
marvinhagemeister, line 140: Wondering if we should be consistent and use `+server.ts` everywhere.

thisisjofrank: yes probably, will update!

marvinhagemeister (approved):
LGTM

Issue #1912, opened 2025-07-16
new theme has huge cumulative layout shift


scarf005:
## Reproduction

https://github.com/user-attachments/assets/5e21f437-9470-43b8-93bc-1682325715fc

URL: https://docs.deno.com/runtime/getting_started/command_line_interface/

tested on macos firefox developer edition 141.0b9 (aarch64)

## Others

new theme ignores [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)

PR #2388, opened 2025-07-21
Shuffle head links to reduce CLS

josh-collinsworth:
Closes https://github.com/denoland/docs/issues/1912

Looks like there are a lot of reasons for the weird blip of loading unstyled content, including network, caching, (I think) Firefox itself, and loading so much CSS from so many different sources. But this reordering of the items in the `<head>` seems to help, at least.

### DIFF
--- a/_includes/layout.tsx
+++ b/_includes/layout.tsx
@@ -18,6 +18,24 @@ export default function Layout(data: Lume.Data) {
         <title>{deleteBackticks(data.title)}</title>
         <link rel="icon" href="/favicon.ico" />
         <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `
+            (function() {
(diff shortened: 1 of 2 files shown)


_includes/layout.tsx
```diff
@@ -18,6 +18,24 @@ export default function Layout(data: Lume.Data) {
         <title>{deleteBackticks(data.title)}</title>
         <link rel="icon" href="/favicon.ico" />
         <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
+        <script
```
philhawksworth, line 21: IN k now this predates this PR, but I'm curious why we use use a call to the `dangerouslySetInnerHTML` method to add javascript to the page instead of... adding the javascript to the page.

This feels inefficient and needlessly fragile to me.

Am I misunderstanding something important here?

josh-collinsworth: May as well fix it while we're here ✅

PR #2400, opened 2025-07-26
fix typo on fmt json

juanpablopiano:
There's a typo in the documentation of the fmt, when copying and pasting this text, it fails because it's semiColons with an 's' at the end.


marvinhagemeister (approved):
Thanks

PR #2424, opened 2025-08-05
Move Nuxt example to EA and update repo links

philhawksworth:
This updates links in the Nuxt tutorial to reflect that the Nuxt example has moved to EA.

- deployed example URL update
- code repo URL update
- Deploy button params update


thisisjofrank (approved):
👍

PR #2425, opened 2025-08-05
Upgrade to Tailwind v4

josh-collinsworth:
**Main change:**
- Upgrades the docs site to Tailwind v4 (🎉)

**All the small changes that make up that big change:**
- Adds some linting rules for Tailwind CSS directives (e.g., `@theme`, `@apply`). No more messy stylesheet squigglies.
- Consolidates **all CSS** into a single `styles.css` file. (This is compiled to a minified output.) This file uses Tailwind layers to sort styles and manage overrides.
- Fixes a few random Markdown issues I came across while doing the work

**What still needs to be done:**
- A whole bunch of those styles in that one mega CSS file can probably be cleaned up now. Since we're using cascase layers, and since we have full control of all the CSS now, we can probably pick off some redundancies and/or move classes into components and markup, instead of having this whole huge mess. But hey, at least the mess is all in one place and fully under our control now.


josh-collinsworth:
> The links no longer seem to have a hover state (except cursor pointer), is this intentional?

Which links do you mean @thisisjofrank ? I went looking but all the links I found behaved the same as on prod.

josh-collinsworth:
> oh interesting, @josh-collinsworth i was particularly looking at the ones on the examples landing page, but perhaps its better that they're now in line with the rest of the site, _do_ you think we should have a hover state?

I modified the hover effect for the links on the examples landing page. It's subtle, but it's there in line with the rest of the site.

PR #2426, opened 2025-08-06
[WIP] Orama search

thisisjofrank:


search.client.ts
```diff
@@ -0,0 +1,696 @@
...
+        if (this.selectedIndex >= 0 && resultsLinks) {
+          const selectedLink =
+            resultsLinks[this.selectedIndex] as HTMLAnchorElement;
+          if (selectedLink) {
+            this.hideResults();
+            // Navigate to the selected result
+            globalThis.location.href = selectedLink.href;
+          }
+        }
+        break;
```
josh-collinsworth, line 201: Do we actually need to handle these cases manually? It seems to me that browser-default tab handling works just as well (if not better), so I worry we're introducing needless complexity here that could result in unforeseen edge cases (particularly since up and down scroll the list when focus is on a link).

Unless there's a reason we need this that native tabbing doesn't handle, I'd say let's remove it.

josh-collinsworth: I guess it is pretty standard, now that I look around a bit.

I was seeing some weird issues with the highlighting, but I think it was actually a VoiceOver issue.

_components/SearchInput.tsx
```diff
@@ -7,7 +7,7 @@ export default function SearchInput() {
           type="search"
           placeholder="Search"
           id="orama-search-input"
-          className="w-full min-w-24 rounded-lg text-sm leading-normal p-1 pl-8 border transition-all duration-150
```
josh-collinsworth: Making the text normal-sized prevents iOS from zooming the input

search.client.ts
```diff
@@ -454,7 +454,7 @@ class OramaSearch {
     );
     return text.replace(
       regex,
-      '<mark class="bg-runtime px-1 py-0.5 rounded text-black font-bold">$1</mark>',
```
josh-collinsworth: Less padding on the sides prevents partial-word highlights from looking goofy (e.g, if the user enters "Types" and gets a hit for "TypeScript")

josh-collinsworth (approved):
I think we may want to continue tweaking our algorithm, but I'm happy with this. Excellent work @thisisjofrank 👏

PR #2431, opened 2025-08-07
Switch hover back to default; minor styles adjustment

josh-collinsworth:
Mostly just restores the default `:hover` behavior, so that hover effects trigger on both fine and coarse pointer devices. Also includes a couple other small CSS fixups.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+  "recommendations": [
+    "denoland.vscode-deno",
+    "bradlc.vscode-tailwindcss"
+  ]
+}


styles.css
```diff
@@ -1,6 +1,6 @@
 @import "tailwindcss";
 
-@theme {
+@theme static {
```
josh-collinsworth, line 3: This ensures our default theme color variables are not tree-shaken. Adds a tiny bit of extra CSS to the final output, but lets you play with all the variables in the browser as needed.

styles.css
```diff
@@ -1017,11 +1020,6 @@
     @apply text-deploy-500 decoration-deploy-500/20 hover:decoration-deploy-500;
   }
 
-  .learn-link {
```
josh-collinsworth, line 1020: These are actually changes that got missed from a separate PR. We don't need the `!`s anymore now that we've got layers, and the `.learn-link` class isn't in the markup anymore (since it's only rendered in one place, so it made more sense to just move these styles there instead of having a separate `@apply` block).

thisisjofrank (approved):
love

_components/ThemeToggle.tsx
```diff
@@ -9,10 +9,12 @@ export default function () {
       <img
         src="/img/light.svg"
         class="w-full block -translate-y-full dark:translate-y-0 transition-transform duration-200 easing-[0.86,0,0.07,1]"
+        alt=""
```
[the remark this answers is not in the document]
josh-collinsworth: There's accessible text inside the button to describe its function already, so I don't think adding descriptions here is probably beneficial. The empty alt just lets the images be ignored.

PR #2432, opened 2025-08-07
Update footer links and styles

josh-collinsworth:
Let's update the footer to look more consistent with the sidebar (and also add a few missing items, and separate the docs into Deno/Services sections).

Also adds a utility class to the theme for the font-sizing of the sidebar and footer, since we're now using it in multiple places.

**Old:**
image

**New:**
image


thisisjofrank (approved):
Perfect!

PR #2438, opened 2025-08-09
Fixed typo (examples/tutorials/fetch_data.md) - changed "bite" to "byte"

geekguy100:
I noticed "bite stream" was used instead of "byte stream", so I decided to open a PR to make the change :)


philhawksworth:
Oops! Thanks @geekguy100

philhawksworth (approved):
Thanks for the catch!

PR #2440, opened 2025-08-11
fix: Command to create a new react app

skeithtan:
The current command creates a Node app, not a Deno app.


bartlomieju:
> The current command creates a Node app, not a Deno app.

What's the difference between the two?

PR #2467, opened 2025-08-19
update accordian to not animate on page load

thisisjofrank:
Its not the most beautiful fix, but it stops the animation


josh-collinsworth:
@thisisjofrank Sorry I missed this earlier. Was there a reason it was closed? Do we have a better solution?

PR #2468, opened 2025-08-19
Feat: Deno KV

arnauorriols:


deploy/early-access/reference/deno-kv.md
```diff
@@ -0,0 +1,96 @@
...
+## Data distribution
+
+Deno KV databases are replicated across at least 3 data centers in the primary
+region, North Virginia (us-east4). Once a write operation is committed, its
+mutations are persistently stored in a quorum of data centers within the primary
+region. Cross-region replication is currently not available.
+
+## Frequently Asked Questions
+
+**Q: What happens to my data when I remove an app assignment?**
```
philhawksworth, line 77: Perhaps we should add a note about what is needed to work locally?

```suggestion
**Q: How is data stored during local development?**

In your local development environment, data is maintained in memory. No database needs to be created or allocated prior to using the KV APIs in local development, and your KV code can be consistent between environments.

**Q: What happens to my data when I remove an app assignment?**
```

PR #2477, opened 2025-08-22
Update open_telemetry.md JS API docs links

sergical:
Current ones go to 404s


philhawksworth (approved):
Most helpful. Thanks!

PR #2491, opened 2025-08-29
fix: remove docs for non-existing lint rule

marvinhagemeister:
The lint rule `button-has-type` doesn't exist. It's called `jsx-button-has-type` for which we already have a page.


philhawksworth (approved):
Good catch. LGTM

PR #2493, opened 2025-09-01
ai search test

thisisjofrank:


josh-collinsworth:
Do we have any way to parse Markdown in the response? It seems like I'm getting a lot of unformatted results in my queries. Maybe not the biggest deal, but kind of annoying. 
CleanShot 2025-09-02 at 14 15 19@2x

josh-collinsworth:
Still not there yet but getting closer. Still need to take a closer look at mobile styles, see if we can transition an actual `<dialog>` element, and make sure focus opens the modal, along with anything else I might have missed. I'll look again in the morning.

josh-collinsworth:
The debouncing generally works well, but I notice in some cases, it doesn't seem like the initial request is properly stopped when the user resumes typing, and you get an odd streaming of two results at once. Here's an example:

https://github.com/user-attachments/assets/0cf5592c-0f4b-4c73-b661-4392d0abd460

PR #2498, opened 2025-09-03
Caching

philhawksworth:
- set default expires header for all assets of 1 hour
- sets expire header of css of 1 day
- adds fingerprint to style.css based on deploy ID to make it immutable and cache bust-able on each deploy


_includes/layout.tsx
```diff
@@ -3,6 +3,10 @@ export function deleteBackticks(str?: string) {
 }
 
 export default function Layout(data: Lume.Data) {
+  const fingerprint = Deno.env.get("DENO_DEPLOY_BUILD_ID") || null;
+
+  console.log("fingerprint", fingerprint);
```
lucacasonato, line 8: log!

PR #2501, opened 2025-09-03
Update Lume

oscarotero:
This PR upgrades the codebase to Lume 3 and introduces some improvements:

- Use folder components for those components that require JavaScript in the browser.
- All `*.client.js` files have been moved to a subdirectory and the `.client` sub-extension has removed since is not needed in Lume 3
- Used Tailwind alone instead of Postcss + Tailwind.
- Lume 3 has [a debugbar](https://lume.land/docs/core/debugbar/) with useful info (you can see it moving the cursor to the bottom of the window). It can be configured to show broken links or any other output. This is what I see in my computer:
  imaxe
  - It looks like the layout rendering is what takes more time. Maybe replacing the .tsx files with a faster format like Vento would help to reduce the time.
  - Saving files takes also a significant amount of time, but I don't know how to optimize this process. It may be an inherent limitation of Deno?

~This PR is in draft because while I was working on this, I discover a couple of bugs in Lume (that's why I'm using a dev version). I'll ship these changes soon and will update this PR with a stable version.~
**Edit:** Lume 3.0.8 is already published

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -9,7 +9,7 @@
     "@std/path": "jsr:@std/path@^1.0.8",
     "@deno/doc": "jsr:@deno/doc@0.169.1",
     "ga4": "https://raw.githubusercontent.com/denoland/ga4/04a1ce209116f158b5ef1658b957bdb109db68ed/mod.ts",
-    "lume/": "https://cdn.jsdelivr.net/gh/lumeland/lume@8773d6d23a426abc55b50ecc1637791f43481724/",
+    "lume/": "https://deno.land/x/lume@v3.0.9/",
     "lume/jsx-runtime": "https://deno.land/x/ssx@v0.1.12/jsx-runtime.ts",
     "googleapis": "npm:googleapis@^144.0.0",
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Setting `compilerOptions.jsx` to `precompile` should noticeably speed up rendering. 

```diff
  // deno.json
  {
    "compilerOptions": {
      "types": [
        "lume/types.ts"
      ],
-     "jsx": "react-jsx",
+     "jsx": "precompile",
      "jsxImportSource": "npm:preact"
    },
  }
```

thisisjofrank:
I've pulled this into #2787 and merged it there, so we can close this one. Thank you so much @oscarotero! I cant believe how much this has cut down the build time! Incredible work!

oscarotero:
Glad to see this finally merged!
Maybe you are interested in upgrading to Lume 3.1.x and use the new `validate_html` plugin to detect HTML errors. Anyway, at least you're using a more modern and faster Lume version :)

PR #2506, opened 2025-09-04
Add docs for permissions in the config file

dsherret:


runtime/fundamentals/security.md
```diff
@@ -95,6 +95,11 @@ By default, Deno will not generate a stack trace for permission requests as it
 comes with a hit to performance. Users can enable stack traces with the
 `DENO_TRACE_PERMISSIONS` environment variable to `1`.
 
+### Configuration File
+
```
bartlomieju, line 99: I think we should call out this is available in 2.5+. CC @philhawksworth @thisisjofrank

PR #2537, opened 2025-09-11
Add paragraph and example for using wildcart for `--allow-net`

csvn:
The docs did not show that it's possible to use `*` as wildcard for subdomains when using hostname for `--allow-net`.

```
deno run --allow-net="*.auth0.com" main.ts
```

Just using "auth0.com" will not allow "<tenant>.eu.auth0.com". To allow any subdomain with `--allow-net`, a wildcard must be used. This proposed change should hopefully make the docs clearer on how to correctly configure permissions.

This is extra important now that it's [possible to configure permissions via `deno.jsonc`](https://deno.com/blog/v2.5#permission-sets-in-config), which makes it easier to make more details settings for permissions:

```jsonc
{
  "permissions": {
    "default": {
      "net": ["*.auth0.com"]
    }
  }
}
```


runtime/fundamentals/security.md
```diff
@@ -219,6 +219,9 @@ Network access is granted using the `--allow-net` flag. This flag can be
 specified with a list of IP addresses or hostnames to allow access to specific
 network addresses.
 
+Hostnames does not allow subdomains, unless explicitly listed. To allow any
```
thisisjofrank, line 222:
```suggestion
Hostnames do not allow subdomains, unless explicitly listed. To allow any
```

thisisjofrank (approved):

PR #2563, opened 2025-09-17
Docs: correct placement of benchmark start

mxdvl:
## Why are we making this change?

Otherwise I assume it’s the same as the previous example. Placing `bench.start()` after the file read is also [what’s done in the “critical sections” examples](https://docs.deno.com/runtime/reference/cli/bench/#critical-sections) of the CLI docs


thisisjofrank (approved):
Thank you

PR #2602, opened 2025-09-26
EA: docs for GitHub repository_dispatch events

lucacasonato:


deploy/early-access/reference/apps.md
```diff
@@ -97,3 +97,67 @@ in the repository dropdown. This will redirect you to GitHub to authorize the
...
+  };
+  revision: {
+    /** The ID of the revision being built. */
+    id: string;
+    /** A URL to view the revision and build status in the Deno Deploy dashboard. */
+    html_url: string;
+    /** The Git commit SHA being built. */
+    git: { sha: string };
+    /** The preview URL the revision is available at, if the build succeeded. */
+    preview_url: string | null;
```
slinkydeveloper, line 142: This is the revision url right? that's great! Will I get this with `deno_deploy.build.routed`? Also why can this be null? when deployment fails?

lucacasonato: This will be the preview URL, yes. It will always be set if the event is `deno_deploy.build.routed`. It will be `null` for all other events.

PR #2618, opened 2025-10-02
move kv files

thisisjofrank:
Move KV docs out of /classic/
Move cron and queues docs into /classic/
Update KV docs wording 
Fix broken images
Update sidebars to match new placement of docs


philhawksworth (approved):
LGTM other than the one issue I have noted in a comment

PR #2642, opened 2025-10-10
feat: add note about root_dir for nvim-lspconfig

martenmatrix:
`nvim-lspconfig` is the official collection of LSP server configurations by NeoVim. Therefore, the majority of NeoVim users use those defaults. It should be included how to prevent attaching `ts_ls` and `denols` to a buffer.

Additionally, [`single_file_support` was renamed to `workspace_required` in NeoVim](https://github.com/neovim/neovim/pull/31824).


bartlomieju (changes requested):
Thanks for the contribution! The `single_file_support` → `workspace_required` update is needed, but there are a few issues to address:

### 1. `root_dir` example is incorrect

The added code block has:
```lua
root_dir = {"package.json"},
```
But `root_dir` expects a function, not a table of strings. It should use `root_pattern` from lspconfig:
```lua
root_dir = require("lspconfig").util.root_pattern("package.json"),
```
As written, this would break users' configs.

### 2. Kickstart.nvim section not updated

The kickstart.nvim example further down the page still uses `single_file_support = false` — this should also be updated to `workspace_required = true` for consistency.

### 3. Fragile GitHub line-number link

The link to `lsp/ts_ls.lua#L56-L68` will break as that file changes. Consider linking to the file without line numbers, or just describing the behavior inline.

lunadogbot (changes requested):
1. `root_dir = {"package.json"}` is still wrong for the `ts_ls` sample. `root_dir` needs a resolver function here; use `require("lspconfig").util.root_pattern("package.json")` or keep this example on the native `root_markers` API.
2. The Kickstart.nvim block still uses `single_file_support = false` for both `ts_ls` and `denols`, so the page now gives conflicting advice after the text switched to `workspace_required = true`. Please update that block too.

- nit: the GitHub link to `lsp/ts_ls.lua#L56-L68` is fragile; link to the file or explain the behavior inline.

fibibot (changes requested):
Two blockers still need fixing before this page can safely recommend the new Neovim config.

1. `root_dir = {"package.json"}` is not a valid `nvim-lspconfig` resolver for `ts_ls`; use `require("lspconfig").util.root_pattern("package.json")`, or keep the example on native `root_markers` instead.
2. The Kickstart.nvim block still says `single_file_support = false` for both `ts_ls` and `denols`, while the text above moved to `workspace_required = true`. Update that block too so the page does not give conflicting configuration.

- nit: the `lsp/ts_ls.lua#L56-L68` link is line-number fragile; link to the file or describe the behavior inline.

fibibot (changes requested):
The Neovim examples now use `workspace_required` and a `root_pattern(...)` resolver consistently. One repo-convention blocker remains:

1. `last_modified: 2026-02-19` still has the old date even though this page now changes the documented Neovim setup. Please bump it for this content update.

martenmatrix:
@bartlomieju bump

PR #2672, opened 2025-10-20
feat: database migrations

arnauorriols:
depends on: https://github.com/denoland/deployng/pull/1708


philhawksworth (approved):
Looks great.

One questions — is it a conscious decision to run the migration _after_ the build rather than _before_?

I'm not an expert here so forgive my ignorance if there are other motives at play, but it might be important for the logic in a build to have access to the new data model which it would not have if this only happened after the build.

PR #2679, opened 2025-10-23
Secondary Indexes recommendations

### LINKED ISSUE #139: KV Secondary Indexes can reference Primary Keys
The current documentation recommends double storing the value of a KvEntry on the creation of a secondary index.  Double storing values requires significant storage and significant effort from the developer to ensure synchronization of all values on all updates. While researching other Deno Deploy documentation, I came across [this Denoland blog entry](https://deno.com/blog/build-crud-api-oak-denokv) by @lambtron and it makes a [completely different recommendation](https://deno.com/blog/build-crud-api-oak-denokv#add-a-secondary-index) of storing the primary key as the value instead of the actual value.  While there's still some onius on the developer to maintain the integrity of the secondary index and using the secondary index is a double lookup, there's no longer multiple copies of the data being stored or multiple transactions when the secondary index is unchanged.  I would like to recommend the current documentation be updated to use primary keys for secondary indices or at least make the suggestion as an exercise for the reader to research.  Please let me know how I can help if either suggestion is determined to have value.

thisisjofrank:
closes #139

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -11,7 +11,7 @@
     "@tailwindcss/cli": "npm:@tailwindcss/cli@^4.1.11",
     "@tailwindcss/postcss": "npm:@tailwindcss/postcss@^4.1.11",
     "ga4": "https://raw.githubusercontent.com/denoland/ga4/04a1ce209116f158b5ef1658b957bdb109db68ed/mod.ts",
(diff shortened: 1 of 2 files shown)


philhawksworth (approved):
Great!

PR #2682, opened 2025-10-23
update the mongoose tutorial to use latest mongoose 

### LINKED ISSUE #84: Deno mongoose does not mention that it does not provide types
https://github.com/denoland/manual/blob/main/node/how_to_with_npm/mongoose.md

There are not types in mongoose library.
So it feels a bit weird using deno with under-the-hood ts support but absence of types for such a big library.

Is there any ways to make deno use npm:mongoose and still provide types?

thisisjofrank:
and strongly typed objects.

Closes #84

### DIFF
--- a/runtime/fundamentals/typescript.md
+++ b/runtime/fundamentals/typescript.md
@@ -11,19 +11,16 @@ oldUrl:
   - /runtime/fundamentals/
 ---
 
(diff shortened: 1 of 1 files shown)


philhawksworth (approved):
LGTM!

PR #2689, opened 2025-10-27
Minor improvement to typescript page

dsherret:
I reviewed this page and made some minor edits. Page looks really good and I don't really have any comments.

### DIFF
--- a/runtime/fundamentals/typescript.md
+++ b/runtime/fundamentals/typescript.md
@@ -11,19 +11,16 @@ oldUrl:
   - /runtime/fundamentals/
 ---
 
-TypeScript is a first class language in Deno, just like JavaScript or
-WebAssembly. You can run or import TypeScript without installing anything more
-than the Deno CLI. With its built-in TypeScript compiler, Deno will compile your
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/typescript.md
```diff
@@ -11,19 +11,16 @@ oldUrl:
...
+config needed. Deno can also type check your TypeScript code, without requiring
+a separate type checking tool like `tsc`.
 
 ## Type Checking
 
 One of the main advantages of TypeScript is that it can make your code type
-safe, catching errors during development rather than runtime. TypeScript is a
-superset of JavaScript meaning that syntactically valid JavaScript becomes
-TypeScript with warnings about being "unsafe".
+safe, catching errors during development rather than runtime.
```
dsherret: Removed "TypeScript is a superset of JavaScript meaning that syntactically valid JavaScript becomes TypeScript with warnings about being "unsafe"." because some people quibble about TS not being a strict superset of JS and also the word "unsafe" makes me personally think about unsafe code in Rust or other languages.

PR #2704, opened 2025-11-04
add a redirect example

### LINKED ISSUE #395: Add documentation on how to setup a redirect.
We should document how to use a redirect to serve a site on both `www.example.com` and `example.com`. 

> One way of redirecting one domain to another is deploying your origin website to the www domain, and setting up another project associated with the non-www domain which serves a simple redirection using JavaScript code as follows:

> `Deno.serve(() => Response.redirect("https://www.example.com/", 301));`

> This code will ensure that visitors to the non-www domain get redirected to the www domain with HTTP status code 301 (permanent redirect).

thisisjofrank:
Closes #395

### DIFF
--- a/api/deno/index.md
+++ b/api/deno/index.md
@@ -555,7 +555,7 @@ console.log(`Result from external addition of 35 and 34: ${result}`);
 Run with `--allow-ffi` and `--unstable` flag:
 
 ```sh
-deno run --allow-ffi --unstable ffi.ts
+deno run --allow-ffi ffi.ts
 ```
 
 ### Non-blocking FFI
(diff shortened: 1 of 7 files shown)


philhawksworth:
I'm not totally convinced this is the best way to do this.

It will indeed redirect any traffic to whatever URL this is served on to www.example.com, but is that desirable? A more desirable pattern would be for some wildcard matching so that we don't just send every request to the home page on a new domain.

so... example.com* -> www.example.com*

I think we should expand this example to work for all urls on the domain and to detect if the www subdomain is omitted.

PR #2734, opened 2025-11-18
fix: Add Prisma Postgres page to menu

arnauorriols:


lunadogbot (changes requested):
This adds an internal sidebar link to , but there is no  or matching directory in this repo. The existing Prisma Postgres material is currently part of , so this would introduce a broken navigation link. Please either point the sidebar item at an existing page/anchor or add the target page in the same PR.

PR #2829, opened 2026-01-13
fix: update sandbox apis

kt3k:
This PR updates the sandbox APIs which are recently updated in upstream.

Main changes
- File system related APIs are now available under `sandbox.fs` namespace
- `.createJsRuntime`, `.repl`, `.eval`, and `.deploy` are moved under `sandbox.deno` namespace
- `lifetime` option has been renamed to `timeout`

closes https://github.com/denoland/deployng/issues/2366

### DIFF
--- a/examples/sandboxes/timeout_control.md
+++ b/examples/sandboxes/timeout_control.md
@@ -37,5 +37,5 @@ await reconnected.kill();
 // At this point, the sandbox is no longer reconnectable
 ```
 
-> Need other lifetime modes? Contact
+> Need other timeout modes? Contact
 > <a href="mailto:deploy@deno.com">deploy@deno.com</a>.


sandboxes/expose_ssh.md
```diff
@@ -12,17 +12,17 @@ import { Sandbox } from "@deno/sandbox";
 
 await using sandbox = await Sandbox.create();
 
-const { hostname, username, privateKey } = await sandbox.exposeSsh();
```
kt3k: exposeSsh does not return a private key. The exposed sandbox can be accessed only using username and hostname

sandboxes/cli.md
```diff
@@ -46,7 +46,7 @@ deno sandbox create --lifetime 2m
 You can also create a sandbox with a custom memory limit:
 
 ```bash
-deno sandbox create --memory 2gb
```
kt3k: This should work soon, but currently there is a bug in backend and this errors. `2gib` works instead.

examples/sandboxes/lifetime_control.md
```diff
@@ -5,13 +5,13 @@ url: /examples/sandboxes_lifetime_control/
 layout: sandbox-example.tsx
```
magurotuna: We may want to update this too?

examples/sandboxes/lifetime_control.md
magurotuna:
```suggestion
// You can still forcibly terminate it before its timeout elapses
```

how about this?

sandboxes/cli.md
magurotuna: lifetime -> timeout

magurotuna: ditto

magurotuna: ditto

kt3k: I created an issue in `@deno/deploy` repo

magurotuna: oh ok that makes sense

PR #2841, opened 2026-01-15
remove old and out of date db tutorials, link to new deploy db instead

thisisjofrank:
None of these tutorials are in date or kept up to date, I think it is better that we remove them entirely to avoid confusion and just link to the new databases feature in deploy.


lunadogbot (changes requested):
Deletions are clean — no internal links to the removed pages, and the sidebar entry in `deploy/classic/_data.ts` is gone with them.

Concern is the redirect targets in `oldurls.json`: `dynamodb`, `firebase`, and arguably `neon-postgres` are being pointed at `/deploy/reference/databases/`, but that page documents Deno Deploy's managed database feature and explicitly says *"Currently, PostgreSQL and Deno KV are supported."* A user landing there from `/deploy/classic/dynamodb/` or `/deploy/classic/firebase/` won't find anything about their database. `/examples/tutorials/connecting_to_databases/` has real Firebase/MongoDB/SQLite content and would serve those entries better; the Postgres/Prisma entries are fine pointing where they do.

- nit: the `"Connecting to databases"` sidebar section in `_data.ts` now contains only Deno KV — worth renaming, or folding the KV entry into `"Deploy platform"`.

fibibot (changes requested):
The redirect map still sends removed third-party database pages to a target that does not cover those databases.

1. `oldurls.json` lines 151-153 map `/deploy/classic/faunadb/`, `/deploy/classic/dynamodb/`, and `/deploy/classic/firebase/` to `/deploy/reference/databases/`, but that page says `Currently, PostgreSQL and Deno KV are supported.` and lists only those engines. Readers following those old URLs will not find FaunaDB, DynamoDB, or Firebase guidance. Keep the existing FaunaDB examples redirect and send DynamoDB/Firebase to `/examples/connecting_to_databases_tutorial/`, or another page that actually covers them.

- nit: `deploy/classic/_data.ts` line 66 now has a `Connecting to databases` section with only Deno KV; fold it into `Deploy platform` or rename the section.

bartlomieju:
Closing this as overtaken by the broader Deno Deploy Classic sunset (#2979). The whole Classic section now carries a sunset banner and is scheduled for shutdown on July 20, 2026, so these five DB tutorials will be removed as part of that teardown rather than piecemeal here. The legacy banners added in #2979 also conflict with this branch. The redirect intent (point old DB tutorial URLs at the new Deploy database docs) is worth folding into the full Classic removal.

PR #2847, opened 2026-01-19
fix: update sandbox CLI page with addition of snapshots and other adjustments

crowlKats:


sandboxes/cli.md
```diff
@@ -238,17 +244,69 @@ deno sandbox volumes list
 You can also search for specific volumes:
 
 ```bash
-deno sandbox volumes list my-data
+deno sandbox volumes list my-volume
 ```
 
 ### Deleting Volumes
```
thisisjofrank: are volumes and snapshots product names or just words? (our headings should be sentence case)

PR #2852, opened 2026-01-21
Tunnel examples

thisisjofrank:
Adding a tutorial for basic use of tunnel, using env vars and adding a database

### DIFF
--- a/examples/tutorials/tunnel_database.md
+++ b/examples/tutorials/tunnel_database.md
@@ -159,7 +159,7 @@ Then, update the HTML to display the notes:
 </ul>
 ```
 
-## Run your app with the tunnel
+## Run your app with tunnel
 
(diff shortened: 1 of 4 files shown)


examples/tutorials/tunnel.md
```diff
@@ -0,0 +1,198 @@
...
+cd svelte-app
+deno run dev
+```
+
+You should now have a Svelte app running locally at `http://localhost:5173` (or
+another port if 5173 is already in use).
+
+## Set up vite, to allow tunneling
+
+To allow the tunnel to work correctly with our Svelte app, we need to make a
```
philhawksworth: Avoid the implication that this is a tunnel limitation

```suggestion
The Vite server used by Svelte is restricted to localhost by default, so to make it more widely available we'll make a  
```

PR #2895, opened 2026-02-09
update sandbox docs as per --help

thisisjofrank:


crowlKats:
maybe adding a link to https://docs.deno.com/sandbox/cli/ would make sense?

PR #2922, opened 2026-02-20
feat: Crons

arnauorriols:
https://docs--crons.deno.deno.net/deploy/reference/crons/


deploy/reference/crons.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Crons
```
philhawksworth: Is there a reason for us calling this feature Crons instead of Cron?

Cron feels more correct, with the descriptive plural form being "cron tasks" or "cron jobs" as we've used in the past and seems to be the wider norm

deploy/reference/crons.md
```diff
@@ -0,0 +1,138 @@
...
+| Every 15 minutes               | `*/15 * * * *` |
+| Every hour                     | `0 * * * *`    |
+| Every 3 hours                  | `0 */3 * * *`  |
+| Daily at 1 AM                  | `0 1 * * *`    |
+| Every Wednesday at midnight    | `0 0 * * WED`  |
+| First of the month at midnight | `0 0 1 * *`    |
+
+### Top-level registration
+
+You must register crons at the module top level, before `Deno.serve()` starts.
```
philhawksworth: People can figure out where to place this code from this section, but it still requires them to do a little thinking. Can we be more explicit about some common or convenient patterns for organising cron tasks into code?

Do people put them in separate files?
Should they / can they be placed in other files as part of other logic?

What's a good pattern for when people decide they are going to type this code somewhere in their project?

deploy/reference/crons.md
```diff
@@ -0,0 +1,138 @@
...
+timeline runs the set of crons defined in its active revision's code. To add,
+remove, or modify crons, update your code and deploy a new revision. Rolling
+back to a previous deployment re-registers the crons from that deployment. You
+can see which crons are currently registered in a given timeline from its page
+in the dashboard.
+
+## Limitations
+
+- Crons must be registered at the module top level (before `Deno.serve()`)
+- No concurrent executions of the same cron
```
philhawksworth: What exactly does that mean? Could we be a tiny bit more descriptive here in the interest of driving out any ambiguity?

philhawksworth (changes requested):
Looking good!

I've added some questions and comments.

A fundamental one is about the naming of this as Crons vs Cron, which might bring about the need for a few edits for consistency depending on what is decided.

deploy/_data.ts
```diff
@@ -54,6 +46,10 @@ export const sidebar = [
         title: "Domains",
         href: "/deploy/reference/domains/",
       },
+      {
+        title: "Crons",
+        href: "/deploy/reference/crons/",
```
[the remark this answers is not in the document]
arnauorriols: Cron jobs?

deploy/reference/crons.md
arnauorriols: Honestly, besides the billing point, the rest are already explained in length above. I'm considering removing the section alltogether, and adding a billing section explaining how billing works

deploy/reference/cron_jobs.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Crons
```
arnauorriols: Renamed feature to "Cron", and normalized all references in text to "cron job(s)". See https://github.com/denoland/docs/pull/2922/changes/e71d5bcec0e39481bd15d8c0d02c684ee583a1b0

PR #2944, opened 2026-03-02
fix: restore copy icon visibility on code block hover

donjo:
## Summary

- The copy icon SVG path was using `fill="hsla(var(--foreground-secondary))"`, referencing a CSS variable (`--foreground-secondary`) that doesn't exist in Tailwind v4 — the correct variable name is `--color-foreground-secondary`
- This made the icon's fill resolve to an invalid/transparent value, so the icon was invisible even though the button itself was present and functional
- Changed `fill` to `currentColor` so the icon inherits the correct color from the button's existing `text-foreground-secondary` CSS class

## Test plan

- [ ] Hover over a code block — copy icon should now be visible in the top-right corner
- [ ] Click the copy icon — text should copy and checkmark should appear
- [ ] Verify in both light and dark mode

### DIFF
--- a/_components/CopyPage.tsx
+++ b/_components/CopyPage.tsx
@@ -1,0 +1,111 @@
+export default function CopyPage({ file }: { file: string | undefined }) {
+  if (!file || file.includes("[")) return null;
+
+  const markdownUrl = `https://docs.deno.com${file}`;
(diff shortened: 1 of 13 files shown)


thisisjofrank (approved):
Ah, good catch!

PR #2950, opened 2026-03-04
docs: document deno install --global --compile flag

bartlomieju:
## Summary
- Document the `--compile` flag for `deno install --global` (Deno 2.7)
- Produces standalone binaries that run without Deno installed
- Added as a new subsection in the install CLI reference page

## Test plan
- [ ] Verify install page renders the new section correctly


thisisjofrank (changes requested):
needs a fmt

PR #2979, opened 2026-03-18
Deno Deploy Classic Sunset

ry:


subhosting/manual/api_migration_guide.md
```diff
@@ -0,0 +1,469 @@
...
+    ]
+  }'
+
+# Create an app that uses the layer
+curl -X POST -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  "https://api.deno.com/v2/apps" \
+  -d '{
+    "slug": "my-service-auth",
+    "layers": ["my-service-shared"],
```
piscisaureus: **Bug:** There is no `stream` query parameter in the v2 API. Streaming is activated by **omitting the `end` parameter**. The v2 OpenAPI spec says for `end`: "Omit for real-time streaming."

Additionally, the `Accept` header controls the streaming format (confirmed in the implementation at `src/routes/api/v2/(_api)/apps.ts`):
- `Accept: text/event-stream` --> SSE
- Otherwise --> NDJSON (`application/x-ndjson`)
- `Accept: application/json` without `end` --> **400 error** with a message telling the user to use SSE/NDJSON or specify `end`

Suggested replacement:
```
**Streaming:** In v1, omitting `since` and `until` enables real-time streaming.
In v2, omit the `end` parameter to stream. Set the `Accept` header to
`text/event-stream` for SSE format or `application/x-ndjson` for NDJSON.
Requesting `application/json` without `end` returns an error.
```

PR #3011, opened 2026-03-30
feat: auto-update last_modified in dev server and add freshness test

bartlomieju:
## Summary
- Adds a `beforeUpdate` event listener in the dev server that automatically updates the `last_modified` frontmatter field to today's date whenever a markdown page is edited during `deno task dev`
- Adds a new test that validates all `last_modified` dates are valid `YYYY-MM-DD`, not in the future, and match the last content-change date in git history (using `git log -G` to skip metadata-only commits)

## Test plan
- [ ] Run `deno test -A frontmatter_test.ts` — all 512 steps should pass
- [ ] Run `deno task dev`, edit a markdown file, and verify the `last_modified` field is updated to today's date
- [ ] Verify no infinite loop: the second watcher trigger should be a no-op since the date already matches

### DIFF
--- a/_config.ts
+++ b/_config.ts
@@ -554,6 +554,42 @@ site.scopedUpdates(
   (path) => path.startsWith("/api/deno/"),
 );
 
+// During dev, auto-update last_modified frontmatter when a page is edited
+site.addEventListener("beforeUpdate", (event) => {
+  const today = new Date().toISOString().slice(0, 10); // YYYY-MM-DD
+
+  for (const file of event.files ?? []) {
+    if (!/\.(md|mdx)$/.test(file)) continue;
+
+    const path = `./${file}`;
+    let content: string;
+    try {
+      content = Deno.readTextFileSync(path);
+    } catch {
+      continue;
+    }
+
+    if (!content.startsWith("---")) continue;
+
+    const endOfFrontmatter = content.indexOf("\n---", 3);
+    if (endOfFrontmatter === -1) continue;
+
+    const frontmatter = content.slice(0, endOfFrontmatter);
+    const rest = content.slice(endOfFrontmatter);
+
+    const lastModifiedMatch = frontmatter.match(/^last_modified:\s*.+$/m);
+    if (lastModifiedMatch) {
+      const updated = frontmatter.replace(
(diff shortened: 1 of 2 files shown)


fibibot (changes requested):
`beforeUpdate` handler in `_config.ts` looks fine. The freshness test in `frontmatter_test.ts` is broken though.

`git log -G "^(?!last_modified:)"` uses PCRE negative lookahead, but git's `-G` is POSIX ERE — on git 2.43 it exits 128 with `fatal: invalid regex: Invalid preceding regular expression` (also fails with `--perl-regexp`, which doesn't apply to `-G`). The `Deno.Command` call doesn't check `result.success` or stderr, so stdout is empty, `gitDates` is empty, and the `if (gitDate)` guard makes `assertEquals(dateStr, gitDate, …)` a no-op for every file. The 220 steps that pass only assert format / parseability / not-in-the-future — the headline "matches last content change" check never runs, and the PR's own "Test plan" 512-step claim doesn't reproduce.

Two things needed:

1. `assert(result.success, ...)` after `outputSync()` so this fails loud if git rejects the regex (otherwise this will silently rot again).
2. Replace the lookahead. e.g. run `git log --name-only --pretty=...` over all commits and filter out `last_modified:`-only commits in-process; or use `--invert-grep --grep` on commit messages if you adopt a marker; the current approach can't be expressed in POSIX ERE.

Holding approval.

lunadogbot (changes requested):
One blocker: the new freshness check can pass without checking any git dates.

1. `"^(?!last_modified:)"` is passed to `git log -G`, but `-G` uses POSIX extended regex and rejects that lookahead with `fatal: invalid regex: Invalid preceding regular expression` on git 2.43. Because `result.success`/stderr are ignored, `gitDates` stays empty and the `if (gitDate)` guard skips every `assertEquals`; `deno test --allow-read --allow-run frontmatter_test.ts` still reports 512 passing steps. Please assert the git command succeeded and replace the lookahead with in-process filtering or another git query that ERE can express.

bartlomieju:
Closing in favor of #3266, which lands the freshness check as a lightweight CI step (diff vs the PR merge-base) instead of the dev-server + git-history approach here.

PR #3013, opened 2026-03-31
docs: expand CPU profiling documentation

bartlomieju:
## Summary
- Added **Customizing profile output** section with practical examples of `--cpu-prof-dir`, `--cpu-prof-name`, and `--cpu-prof-interval` flags
- Added **Profiling tips** section covering representative workloads, self time vs. total time, before/after comparisons, combining output formats, and filtering startup noise

## Test plan
- [ ] Verify the page renders correctly on the docs site
- [ ] Confirm all code examples are syntactically correct

### DIFF
--- a/runtime/fundamentals/debugging.md
+++ b/runtime/fundamentals/debugging.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-03-19
+last_modified: 2026-05-20
 title: "Debugging"
 description: "Complete guide to debugging Deno applications. Learn to use Chrome DevTools, VS Code debugger, and other debugging techniques for TypeScript/JavaScript code in Deno."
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
Lint CI is red — `deno fmt --check` fails on the new "Profiling tips" bullets in `runtime/fundamentals/debugging.md` (lines 330–349). The bullets wrap at the wrong column; running `deno fmt runtime/fundamentals/debugging.md` re-wraps them and clears the check.

Content is consistent with the flags table at lines 199–204: defaults (`1000μs` / `1ms`), `--cpu-prof-dir` defaulting to cwd, and the `CPU.<timestamp>.<pid>.cpuprofile` naming all match.

PR #3020, opened 2026-04-01
change main docs structuring

crowlKats:
image (4)
image (3)
image (2)
image (1)

### DIFF
--- a/_components/Navigation.tsx
+++ b/_components/Navigation.tsx
@@ -73,5 +73,22 @@ function getSectionData(data: Lume.Data, currentUrl: string) {
 
   // Fall back to the default behavior using just the first segment
   const dataPath = urlSegments[0];
-  return data.search.data(`/${dataPath}/`)?.sidebar;
+  const sectionData = data.search.data(`/${dataPath}/`);
+
+  // If the section has multi-sidebar routing (e.g. runtime), pick the right one
+  const sidebarUrlMap = sectionData?.sidebarUrlMap as
+    | Record<string, string>
+    | undefined;
+  const sidebars = sectionData?.sidebars as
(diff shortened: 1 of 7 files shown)


lunadogbot (changes requested):
1. `sidebarUrlMap[normalizedUrl]` only does an exact lookup, and `runtime/_data.ts` now keeps `Standard Library` as a single `/runtime/reference/std/` entry. Existing nested std pages such as `/runtime/reference/std/assert/` are no longer mapped to the package sidebar, so they fall back to `basicsSidebar`. Please add prefix routing for this tree or keep/generated-map the std module hrefs.
2. `reference_gen/parallel-doc.ts` is under the generated reference tree. Repo rules keep `reference/` and `reference_gen/` changes upstream-generated, so this formatting fix should come from the generator/upstream source or be split with maintainer confirmation.

fibibot (changes requested):
The runtime subnav routing loses nested pages that are not enumerated in `sidebarUrlMap`, and this also hand-edits generated reference output.

1. `sidebarUrlMap[normalizedUrl]` only accepts exact URLs, but `runtime/_data.ts` now keeps std as a single `Standard Library` entry at `/runtime/reference/std/`. Pages like `/runtime/reference/std/assert/` are not collected, so they fall back to `basicsSidebar` instead of the package/std sidebar. Add prefix routing for std or keep the child hrefs in the map.
2. `reference_gen/parallel-doc.ts` is under generated reference output. Repo policy keeps `reference/` and `reference_gen/` changes generated from upstream sources, so this formatting fix should be split or regenerated instead of hand-edited here.

bartlomieju:
Closing as stale and I'm iterating on this one right now as well

PR #3026, opened 2026-04-08
docs: add tsconfig.json support details to TypeScript page

### LINKED ISSUE #2666: Details of tsconfig support
Add to https://docs.deno.com/runtime/fundamentals/typescript/ to explain:

- Describe the levels of tsconfig support
- include examples of configuring with tsconfig

Since Deno 2.1 we autodetect jsconfig when a package.json is present

bartlomieju:
Adds a short "Configuring TypeScript with `tsconfig.json`" section to the TypeScript fundamentals page, explaining that Deno auto-detects `tsconfig.json` in any workspace directory that contains a `deno.json` or `package.json`, and (since Deno 2.1) auto-detects `jsconfig.json` when a `package.json` is present. Includes a short example of an existing Node.js `tsconfig.json` being picked up unchanged, and links to the existing `ts_config_migration` reference for the full list of supported fields, precedence rules, and compiler option defaults rather than duplicating that content here.

Closes #2666

### DIFF
--- a/runtime/fundamentals/typescript.md
+++ b/runtime/fundamentals/typescript.md
@@ -264,6 +264,46 @@ To specify the library files to use in a TypeScript file, you can use
 /// <reference lib="dom" />
 ```
 
+## Configuring TypeScript with `tsconfig.json` {#tsconfig}
+
+While Deno uses `deno.json` for TypeScript configuration by default, it also
+supports `tsconfig.json` for compatibility with existing Node.js and TypeScript
+projects, making it easier to adopt Deno incrementally. Each workspace
+directory containing a `deno.json` or `package.json` is probed for a
+`tsconfig.json` — if one exists, Deno will automatically use it for type
+checking and the language server, no extra flags needed. Since Deno 2.1,
+`jsconfig.json` is also auto-detected when a `package.json` is present, which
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
Blockers:

1. CI is red because `deno fmt --check` fails on this section — lines 314–315 (`- Among tsconfig references, a more specific path…`) and 343–344 (`[configuring TypeScript](…) reference for a full list…`) need reflowing. `deno fmt runtime/fundamentals/typescript.md` fixes it.
2. "Since Deno 2.1, `jsconfig.json` files are also auto-detected when a `package.json` is present." — I couldn't find a canonical source for the 2.1 version claim (the existing `runtime/reference/ts_config_migration.md` doesn't mention `jsconfig.json` at all). Can you link to the Deno PR or release notes that added this, or soften the version assertion?

Content otherwise mirrors `runtime/reference/ts_config_migration.md` lines 98–138 — same auto-detection paragraph, same supported-fields block, same precedence rules. Fine to surface this on the fundamentals page, but the two pages will need to be kept in sync going forward.

- nit: precedence list drops one rule from the reference page ("A reference takes precedence over its referrer.") — intentional simplification or oversight?

Issue #2989, opened 2026-03-21
Feedback: /examples/configuration_with_deno_json/ - Needs Improvement


denobot:
## Documentation Feedback

**Path:** [/examples/configuration_with_deno_json/](https://docs.deno.com/examples/configuration_with_deno_json/)
**Helpful:** No ❌

**Feedback:**
> What is the recommended separator for multi-word tasks, dash or colon?

E.g. "tasks.test:stress" vs. "tasks.test-stress".

---
*This issue was automatically created from the documentation feedback system.*

PR #3028, opened 2026-04-08
docs: use colons as task name separators in examples

bartlomieju:
Updates the task naming examples in `runtime/reference/cli/task.md` to use `:` as the separator (e.g. `build:client`, `dev:server`) instead of `-`, bringing this page in line with the rest of the docs, which already use the colon convention (`test:all`, `build:full`, `dev:api`, `lint:fix`, etc.). Adds a short note explicitly recommending `:` for multi-word task names — answering the question raised in #2989 — and incidentally fixes a preexisting "sever" typo in the dev-server example.

Ref #2989

### DIFF
--- a/runtime/reference/cli/task.md
+++ b/runtime/reference/cli/task.md
@@ -88,18 +88,22 @@ pattern. A wildcard pattern is specified with the `*` character.
 ```json title="deno.json"
 {
   "tasks": {
-    "build-client": "deno run -RW client/build.ts",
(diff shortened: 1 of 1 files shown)


lunadogbot (approved):
Mechanical rename to colon-separated task names; CI green.

- nit: line 214 still has `sever/mod.ts` — pre-existing typo, but you're touching that exact line so it's easy to fix in this PR.

PR #3029, opened 2026-04-08
docs: simplify file existence example to use node:fs

### LINKED ISSUE #2987: Feedback: /examples/checking_file_existence/ - Needs Improvement
## Documentation Feedback

**Path:** [/examples/checking_file_existence/](https://docs.deno.com/examples/checking_file_existence/)
**Helpful:** No ❌

**Feedback:**
> I just need file.exists function as primary citizen. Because checking file existence is a primary business logic in many cases. For example:
if (!fa.exists('git/repo/path')) {
   throw new Error('You should clone git repo first to start test stand')
}

---

bartlomieju:
## Summary
- Rewrites the file existence checking example to use `node:fs` `exists` and `node:fs/promises` `access` APIs
- Removes the Deno-specific `@std/fs/exists` and `Deno.lstat` approaches in favor of the simpler, more familiar Node.js-compatible APIs

Closes #2987

## Test plan
- [ ] Verify the example renders correctly on the docs site
- [ ] Run the example script with `deno run -R`

### DIFF
--- a/examples/scripts/checking_file_existence.ts
+++ b/examples/scripts/checking_file_existence.ts
@@ -2,33 +2,26 @@
  * @title Checking for file existence
  * @difficulty beginner
  * @tags cli, deploy
- * @run -R -W <url>
+ * @run -R <url>
  * @group File System
  *
  * When creating files it can be useful to first ensure that
  * such a file doesn't already exist.
- * There are a number of ways to do this.
  */
 
-// Use the `exists` utility from the std library to check for existence of a file or folder.
-// Note: Can create a race condition if followed by file operation.
-// Consider the alternative below.
-import { exists } from "jsr:@std/fs/exists";
-await exists("./this_file_or_folder_exists"); // true
-await exists("./this_file_or_folder_does_not_exist"); // false
+// Use the Node.js-compatible `fs.exists` function for a simple check.
+import { exists } from "node:fs";
 
-// We can also use this function to check if the item on a path is a file or a directory
-await exists("./file", { isFile: true }); // true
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
Two issues to address:

1. `fs.exists` is [deprecated upstream](https://nodejs.org/api/fs.html#fsexistspath-callback) — Node's fs docs mark it "Stability: 0 — Deprecated: Use fs.stat() or fs.access() instead." A beginner-tagged example shouldn't lead with a deprecated callback API. The promise-based `access` block is the recommended replacement; consider making it the only example.

2. The TOCTOU warning was deleted. The original called out "Can create a race condition if followed by file operation" and pointed at the `Deno.lstat` pattern as the race-free alternative. The new example removes the warning — and the linked feedback (#2987) is asking to use this for `if (!exists(path)) { create() }`, which is exactly the racy pattern users need to be steered away from.

lunadogbot (changes requested):
1. `import { exists } from "node:fs"` puts a deprecated callback API into a beginner-tagged example. Node marks `fs.exists()` as deprecated and recommends `fs.stat()` or `fs.access()` instead (https://nodejs.org/api/fs.html#fsexistspath-callback), so this page should keep the promise-based `access` example or use direct operation+error handling.
2. The race-condition warning was removed. The intro still frames this as useful before creating files, but checking existence before a later write is exactly the TOCTOU pattern the old example warned about; keep that warning or rewrite the example so it handles `EEXIST`/`NotFound` at the operation site.

PR #3032, opened 2026-04-08
docs: add file reading example to CSV parsing page

### LINKED ISSUE #2894: Feedback: /examples/parsing_serializing_csv/ - Needs Improvement
## Documentation Feedback

**Path:** [/examples/parsing_serializing_csv/](https://docs.deno.com/examples/parsing_serializing_csv/)
**Helpful:** No ❌

**Feedback:**
> I did find what I needed but the information provided is incomplete. For example: what if my data is in a CSV file, instead of being hard-coded like in the example? How do I access it? Maybe you cover this in another article, but where's the link to it?

---

bartlomieju:
Adds an example showing how to read and parse a CSV file with `Deno.readTextFile` at the top of the page, addressing feedback that the existing example only covered hardcoded strings. The original string-based examples are kept below for reference, and `@run` gains `-R` so the example runs with read permission.

Closes #2894

### DIFF
--- a/examples/scripts/parsing_serializing_csv.ts
+++ b/examples/scripts/parsing_serializing_csv.ts
@@ -2,7 +2,7 @@
  * @title Parsing and serializing CSV
  * @difficulty beginner
  * @tags cli, deploy, web
- * @run <url>
+ * @run -R <url>
  * @resource {/examples/import_export} Example: Importing & Exporting
  * @resource {https://datatracker.ietf.org/doc/html/rfc4180} Spec: CSV
  * @group Encoding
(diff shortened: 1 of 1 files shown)


lunadogbot (approved):
Addresses the feedback in #2894 — readers wanted exactly the file-reading example this PR adds. `@run -R` matches the new `Deno.readTextFile("data.csv")` call and follows the convention in `examples/scripts/watching_files.ts` and `hashing.ts`. The reworded inline-string comment ("you can also parse CSV strings directly") reads coherently with the new file-first ordering. CI green.

PR #3037, opened 2026-04-08
docs: fix static module graph description to include all module sources

### LINKED ISSUE #2686: Feedback: /runtime/fundamentals/security/ - Needs Improvement
## Documentation Feedback

**Path:** /runtime/fundamentals/security/
**Helpful:** No ❌

**Feedback:**
> In the key principles section, the last point states that deno will load local files in the static module graph without restrictions. That doesn't apply only to local files though, deno will also load npm modules, jsr modules, and a few other sources without requesting permissions (this is noted later on in `Importing from the web`

**GitHub User:** @nathanwhit

---

bartlomieju:
## Summary
- Fixes the "key principles" bullet in the security page that incorrectly stated only local files can be imported without restrictions in the static module graph
- Updates wording to clarify that npm packages, jsr packages, and remote URLs are also loaded without restrictions in the initial static module graph

Closes #2686

## Test plan
- [ ] Verify the updated bullet point renders correctly

### DIFF
--- a/runtime/fundamentals/security.md
+++ b/runtime/fundamentals/security.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-03-26
+last_modified: 2026-05-21
 title: "Security and permissions"
 description: "A guide to Deno's security model and permissions system. Learn about secure defaults, permission flags, runtime prompts, and how to safely execute code with granular access controls."
 oldUrl:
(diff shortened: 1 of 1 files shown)


fibibot (approved):
Wording fix is correct — Deno's initial static module graph loads npm, jsr, and remote modules without runtime permission checks, the same as local files. The original bullet implied this was a local-file-only carve-out, which was misleading.

PR #3038, opened 2026-04-08
docs: clarify that ignore skips evaluation and type-checking

### LINKED ISSUE #2657: Feedback: /runtime/reference/documentation/ - Needs Improvement
## Documentation Feedback

**Path:** /runtime/reference/documentation/
**Helpful:** No ❌

**Feedback:**
> Text about `ignore` attribute for code snippets doesn't mention it skipping evaluation.

Says "Another attribute supported is ignore, which tells the test runner to skip type-checking the code block."

I believe it should say "Another attribute supported is ignore, which tells the test runner to skip evaluation and type-checking the code block."

**GitHub User:** @iuioiua

---

bartlomieju:
## Summary
- Fixes the description of the `ignore` attribute for doc test code blocks to mention it skips both evaluation and type-checking, not just type-checking

Closes #2657

## Test plan
- [ ] Verify the updated text renders correctly

### DIFF
--- a/runtime/reference/documentation.md
+++ b/runtime/reference/documentation.md
@@ -41,7 +41,7 @@ If no language identifier is specified then the language is inferred from media
 type of the source document that the code block is extracted from.
 
 Another attribute supported is `ignore`, which tells the test runner to skip
-type-checking the code block.
+evaluation and type-checking of the code block.
 
 ````ts
 /**


fibibot (approved):
Correct clarification — `ignore` blocks are skipped entirely by `deno test --doc`, not just untyped. The existing example at `runtime/reference/documentation.md:46` (`const x: string = 42;`) only has a type error, so the previous wording wasn't visibly wrong, but the new wording matches actual behaviour.

PR #3041, opened 2026-04-08
docs: add xml module to std menu

tomas-zijdemans:


lunadogbot (approved):
The page at `/runtime/reference/std/xml/` is auto-generated from JSR's `@std` scope listing, where `xml` is present, so the sidebar entry will resolve. Alphabetical placement between `webgpu` and `yaml` is correct. Lint+link CI is green.

- nit: separate from this PR, `math` is also in the JSR `@std` listing but missing from this sidebar array — worth a follow-up.

fibibot (approved):
`runtime/_data.ts` adds `xml` in the std package sidebar between `webgpu` and `yaml`. `deno info jsr:@std/xml` resolves the current package, GitHub CI is green, and `deno task build:light` also completed successfully locally.

PR #3051, opened 2026-04-16
docs: fix outdated OpenTelemetry api example

RYGRIT:
## Problem

`trace.SpanStatusCode.ERROR` is not available in @opentelemetry/api 1.x.

```bash
deno check server.ts

TS2339 [ERROR]: Property 'SpanStatusCode' does not exist on type 'TraceAPI'.
          code: trace.SpanStatusCode.ERROR,
                      ~~~~~~~~~~~~~~
```

## Changes

- Import `SpanStatusCode` directly from `@opentelemetry/api`
- Replace `trace.SpanStatusCode.ERROR` with `SpanStatusCode.ERROR`

### DIFF
--- a/examples/tutorials/basic_opentelemetry.md
+++ b/examples/tutorials/basic_opentelemetry.md
@@ -22,7 +22,7 @@ Let's start by creating a basic HTTP server that simulates a small web
 application:
 
 ```ts title="server.ts"
-import { metrics, trace } from "npm:@opentelemetry/api@1";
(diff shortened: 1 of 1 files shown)


lunadogbot (approved):
Correct fix — `SpanStatusCode` is a top-level export of `@opentelemetry/api`, not a property of `trace`, so `trace.SpanStatusCode.ERROR` would fail type-check against the pinned `@opentelemetry/api@1`. CI green.

PR #3059, opened 2026-04-23
Fix bundle command syntax in documentation

nicholasq:
I believe the current deno docs are incorrect for the first code block under [Basic usage](https://docs.deno.com/runtime/reference/cli/bundle/#basic-usage)

It shows:
```bash
deno bundle main.ts output.js
```
Which does not work for me on deno 2.7.5. I believe the above is missing the `-o` output flag. It should be something like this:
```bash
deno bundle -o output.js main.ts
```

### DIFF
--- a/runtime/reference/cli/bundle.md
+++ b/runtime/reference/cli/bundle.md
@@ -14,7 +14,7 @@ JavaScript file.
 ## Basic usage
 
 ```sh
-deno bundle main.ts output.js
+deno bundle -o output.js main.ts
 ```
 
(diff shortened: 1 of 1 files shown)


lunadogbot (approved):
Original `deno bundle main.ts output.js` (two positional args) doesn't work — the `-o` form here matches the canonical example at `bundling.md:46`. CI green.

Duplicate: #3105 proposes the same fix using a `>` redirect (the other form shown at `bundling.md:42`). Either works; flagging so you can dedupe.

PR #3064, opened 2026-04-29
docs: document delta updates for `deno upgrade` (2.8)

lunadogbot:
## Summary

Documents the new bsdiff delta-update mechanism shipping in Deno 2.8 ([denoland/deno#33274](https://github.com/denoland/deno/pull/33274)).

- Adds a "Delta updates" section to `runtime/reference/cli/upgrade.md` describing automatic patch-based upgrades.
- Documents the new `--no-delta` flag for forcing full-archive downloads.
- Notes that delta patches and the resulting binary are SHA-256 verified before installation.

## Test plan

- [x] `deno task serve` renders the new section without layout regressions.

### DIFF
--- a/runtime/reference/cli/upgrade.md
+++ b/runtime/reference/cli/upgrade.md
@@ -88,6 +88,27 @@ on GitHub:
 curl -sL https://github.com/denoland/deno/releases/download/v2.7.0/deno-x86_64-unknown-linux-gnu.zip.sha256sum
 ```
 
+## Delta updates
+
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `so the upgrade always succeeds` overstates the fallback behavior. The implementation only returns `None` from the delta path and continues through the normal full-archive download; that path can still fail on network errors, missing release assets, checksum mismatch, or install errors. Please change this to something like “so delta failures fall back to the full archive path.”

PR #3065, opened 2026-04-29
docs: mark `no-process-global` and `no-node-globals` as opt-in (2.8)

lunadogbot:
## Summary

Documents the lint default change shipping in Deno 2.8 ([denoland/deno#33247](https://github.com/denoland/deno/pull/33247)). `deno_lint` 0.84.0 removes `no-process-global` and `no-node-globals` from the recommended set, so they are off by default.

- Drops the `recommended` tag from both rule pages.
- Adds a short opt-in `deno.json` snippet to each rule page.
- Tweaks `runtime/fundamentals/node.md` to match — the warning shown in the example is now opt-in, not default.

## Test plan

- [x] `deno task serve` renders the two rule pages and the Node.js page without layout regressions.
- [ ] Confirm with reviewers whether the lint index (`/lint/`) needs the rule list adjusted — these rules will no longer appear under the "recommended" filter, which is the intent.

### DIFF
--- a/lint/rules/no-node-globals.md
+++ b/lint/rules/no-node-globals.md
@@ -1,5 +1,5 @@
 ---
-tags: [recommended]
+tags: []
 ---
 
 Disallows the use of NodeJS global objects.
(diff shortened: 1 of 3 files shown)


fibibot (changes requested):
Please bump `last_modified` in `runtime/fundamentals/node.md` for the Node compatibility guide change. That page currently remains at `2026-02-10` on the 2.8 base. The lint rule pages in this PR use their existing minimal frontmatter pattern, so I’m only calling out the runtime docs page.

PR #3066, opened 2026-04-29
docs: add `deno bump-version` reference page (2.8)

lunadogbot:
## Summary

Adds a CLI reference page for the new `deno bump-version` subcommand shipping in Deno 2.8 ([denoland/deno#30562](https://github.com/denoland/deno/pull/30562)).

The command updates the `version` field in `deno.json(c)` or `package.json` and accepts the same seven semver-style increments as `npm version` (`patch`, `minor`, `major`, `prepatch`, `preminor`, `premajor`, `prerelease`).

- New page at `runtime/reference/cli/bump_version.md` with usage and an examples table.
- Entry added to the CLI sidebar (`runtime/_data.ts`) and to the CLI index page.

## Test plan

- [x] `deno task serve` — page renders, sidebar entry resolves.
- [ ] After 2.8 ships, the auto-generated CLI flag table will pick up `bump-version`; we can extend the page then if any new flags are added.

### DIFF
--- a/runtime/_data.ts
+++ b/runtime/_data.ts
@@ -113,6 +113,10 @@ export const sidebar = [
             href: "/runtime/reference/cli/bench/",
           },
           {
+            title: "deno bump-version",
+            href: "/runtime/reference/cli/bump_version/",
+          },
+          {
             title: "deno bundle",
             href: "/runtime/reference/cli/bundle/",
           },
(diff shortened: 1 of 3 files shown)


fibibot (changes requested):
There is a factual mismatch with the merged Deno implementation. This page says the command uses the first `version` field found in either `deno.json(c)` or `package.json`, but denoland/deno#30562 was updated before merge to prioritize `deno.json` over `package.json` (see the final commit note: "prioritize deno.json over package.json"). Please update that sentence so readers know which file wins when both exist.

PR #3070, opened 2026-04-29
docs: add `deno pack` reference page (2.8)

lunadogbot:
## Summary

Adds a CLI reference page for the new `deno pack` command shipping in Deno 2.8 ([denoland/deno#32139](https://github.com/denoland/deno/pull/32139)). The command builds an npm-compatible tarball from a Deno project so libraries can be published to the npm registry alongside JSR.

- New page at `runtime/reference/cli/pack.md` with usage, npm-publish workflow, and a "when to use" comparison vs `deno publish`.
- Sidebar entry in `runtime/_data.ts`, entry in the CLI index page.

## Test plan

- [x] `deno task serve` — page renders, sidebar entry resolves.
- [ ] If the upstream PR adds flags before merge (e.g. `--out-dir`, `--dry-run`), this page will need a flags section.

### DIFF
--- a/runtime/_data.ts
+++ b/runtime/_data.ts
@@ -181,6 +181,10 @@ export const sidebar = [
             href: "/runtime/reference/cli/outdated/",
           },
           {
+            title: "deno pack",
+            href: "/runtime/reference/cli/pack/",
+          },
+          {
             title: "deno publish",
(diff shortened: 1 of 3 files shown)


ejsmith:
Does this replace DNT? I can't tell if I should switch over to this from using DNT.

PR #3077, opened 2026-04-29
docs: document npm tarball installs (2.8)

lunadogbot:
## Summary

Documents the new ability to install npm packages directly from local or remote tarballs in Deno 2.8 ([denoland/deno#32945](https://github.com/denoland/deno/pull/32945)).

- New "Installing from npm tarballs" section in `runtime/reference/cli/install.md` with examples for local file, npm-registry URL, and GitHub URL.
- Documents detection rules (`.tgz`/`.tar.gz` local; any non-`git+` http URL remote), transitive resolution behavior, lockfile recording, and the local cache directory `.deno_tarball_cache/`.
- Calls out the `package.json`-only limitation as an info admonition.

## Test plan

- [x] `deno task serve` — section renders, examples readable.

### DIFF
--- a/runtime/reference/cli/install.md
+++ b/runtime/reference/cli/install.md
@@ -46,6 +46,42 @@ If your project has a `package.json` file, the packages coming from npm will be
 added to `dependencies` in `package.json`. Otherwise all packages will be added
 to `deno.json`.
 
+### Installing from npm tarballs
+
+Starting in Deno 2.8, `deno install` accepts npm tarballs directly — both local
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `Starting in Deno 2.8, deno install accepts npm tarballs directly` documents behavior from denoland/deno#32945, but that runtime PR is still OPEN. Until the implementation lands in the 2.8 branch/current canary, this page would publish a CLI feature that users cannot run yet; please land the runtime change first or mark this PR as blocked on it.

- nit: `last_modified: 2026-03-12` should be bumped for this content addition.

PR #3081, opened 2026-04-29
docs: document `NODE_EXTRA_CA_CERTS` (2.8)

lunadogbot:
## Summary

Adds `NODE_EXTRA_CA_CERTS` to the Special environment variables table on `runtime/reference/env_variables.md`, documenting the new Node-compat env var supported in Deno 2.8 ([denoland/deno#33148](https://github.com/denoland/deno/pull/33148)).

- Loaded at the root certificate store level, so it applies to native `fetch()`, `Deno.connectTls()`, and the Node polyfills `node:https` / `node:tls`.
- Missing or invalid PEM files emit a warning (no hard error) — matches Node.js semantics.

## Test plan

- [x] `deno task serve` — table renders, the new row aligns.


fibibot (changes requested):
1. `last_modified: 2025-12-19` is unchanged even though the PR adds a new documented runtime environment variable. Please bump the page frontmatter to the date this page is being updated so the freshness metadata stays accurate.

PR #3082, opened 2026-04-29
docs: document `deno x --package` flag (2.8)

lunadogbot:
## Summary

Documents the new `--package` / `-p` flag on `deno x` shipping in Deno 2.8 ([denoland/deno#32855](https://github.com/denoland/deno/pull/32855)). It separates the package to install from the binary to execute, matching the `npx -p <package> <binary>` convention.

- New "Specifying the package separately from the binary" section in `runtime/reference/cli/x.md`.
- Examples for both `-p typescript tsc` and the version-pinned `-p typescript@5 tsc`.
- Notes that the existing `package/binary` form is unaffected.

## Test plan

- [x] `deno task serve` — section renders, anchor resolves.

### DIFF
--- a/runtime/reference/cli/x.md
+++ b/runtime/reference/cli/x.md
@@ -45,6 +45,23 @@ Run a JSR package:
 deno x jsr:@std/http/file-server
 ```
 
+## Specifying the package separately from the binary
+
+Some npm packages expose multiple binaries — `typescript` ships both `tsc` and
+`tsserver`, for example. Starting in Deno 2.8, `--package` (`-p`) lets you
+choose the package and the binary independently, matching the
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `command: x` means this page renders the generated options table from `_commands_reference.json`, but that data still only lists `--check`, `--env-file`, `--install-alias`, and `--unstable` for `deno x`. The new prose documents `--package`, but the generated options section will still omit it; regenerate/update the CLI reference data from the Deno 2.8 CLI so the flag appears in the command reference itself.

- nit: `last_modified: 2025-12-10` should be bumped for this content addition.

bartlomieju (approved):

PR #3084, opened 2026-04-29
docs: document `OffscreenCanvas` (2.8)

lunadogbot:
## Summary

Documents the new `OffscreenCanvas` web API shipping in Deno 2.8 ([denoland/deno#29357](https://github.com/denoland/deno/pull/29357)).

- New "OffscreenCanvas" section in `runtime/reference/web_platform_apis.md`.
- Includes a 2D-drawing + `convertToBlob` example and a list of typical use cases (server-side image generation, Web Worker off-thread drawing, thumbnails).
- Notes the WebGL/WebGPU-on-OffscreenCanvas limitation.

## Test plan

- [x] `deno task serve` — section renders ahead of the "Deviations of other APIs from spec" heading.
- [ ] If the upstream PR's spec coverage changes (e.g. adds WebGL2 support) before merge, the limitation note should be updated.

### DIFF
--- a/runtime/reference/web_platform_apis.md
+++ b/runtime/reference/web_platform_apis.md
@@ -489,6 +489,34 @@ const worker = new Worker(import.meta.resolve("./worker.js"), {
 });
 ```
 
+## OffscreenCanvas
+
+Starting in Deno 2.8, the
+[`OffscreenCanvas`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)
+API is available. `OffscreenCanvas` is a canvas that lives outside any DOM and
+can be used in any context — including Web Workers — for off-thread rendering
+and image generation.
+
+```ts
+const canvas = new OffscreenCanvas(256, 256);
+const ctx = canvas.getContext("2d");
+ctx.fillStyle = "tomato";
+ctx.fillRect(0, 0, 256, 256);
+
+// Encode to PNG / JPEG / WebP via convertToBlob()
+const blob = await canvas.convertToBlob({ type: "image/png" });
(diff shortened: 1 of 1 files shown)


runtime/reference/web_platform_apis.md
```diff
@@ -489,6 +489,34 @@ const worker = new Worker(import.meta.resolve("./worker.js"), {
...
+[`OffscreenCanvas`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)
+API is available. `OffscreenCanvas` is a canvas that lives outside any DOM and
+can be used in any context — including Web Workers — for off-thread rendering
+and image generation.
+
+```ts
+const canvas = new OffscreenCanvas(256, 256);
+const ctx = canvas.getContext("2d");
+ctx.fillStyle = "tomato";
+ctx.fillRect(0, 0, 256, 256);
```
Hajime-san: @crowlKats 
I think its wrong. It's only `bitmaprenderer` or `webgpu` for now.

crowlKats: yeah this bot is assuming a lot of things for these docs that is wrong. i will revisit these in the coming days, before the release

lunadogbot: Good catch — fixed in 657f1d8 / 4a01094. The section now lists only `bitmaprenderer` and `webgpu` as supported, and explicitly notes that `2d`, `webgl`, and `webgl2` return `null`. The example also uses `bitmaprenderer` rather than 2D drawing.

crowlKats (approved):

PR #3086, opened 2026-04-29
docs: update first-project guide for new `deno init` template (2.8)

lunadogbot:
## Summary

Updates the "Making a Deno project" walkthrough to match the new default `deno init` template shipping in Deno 2.8 ([denoland/deno#33042](https://github.com/denoland/deno/pull/33042)). The template now scaffolds a small `Deno.serve` HTTP server (with two tests) instead of the old `add(2, 3)` math example.

- Rewrites the description of `main.ts` and the run/test output blocks in `runtime/getting_started/first_project.md`.
- Mentions the `import.meta.main` guard so readers understand why the tests import the handler directly.

## Test plan

- [x] `deno task serve` — page renders.
- [ ] If reviewers want me to verify the exact test names against the upstream template, I can adjust before merge.

### DIFF
--- a/runtime/getting_started/first_project.md
+++ b/runtime/getting_started/first_project.md
@@ -35,11 +35,12 @@ my_project
 
 A `deno.json` file is created to
 [configure your project](/runtime/fundamentals/configuration/), and two
-TypeScript files are created; `main.ts` and `main_test.ts`. The `main.ts` file
-is where you'll write your application code, on initial creation it will contain
-a simple program which adds two numbers together. The `main_test.ts` file is
-where you can write tests, initially it will contain a test for your addition
-program.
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
Two generated-template details do not match merged denoland/deno#33042, so readers would copy commands/output that the 2.8 `deno init` template does not produce.

1. `$ deno main.ts` should be `deno run --allow-net main.ts`; upstream init prints that command, and the scaffolded `Deno.serve(handler)` path needs net permission.
2. `handler returns hello` / `handler returns 404 for unknown route` should match the generated `main_test.ts` names: `returns html on /` and `returns json on /api`. The template does not generate a 404 test.

- nit: bump `last_modified`; this is a substantive content update.

PR #3087, opened 2026-04-29
docs: document `deno upgrade pr` (2.8)

lunadogbot:
## Summary

Documents the new `deno upgrade pr <number>` subcommand shipping in Deno 2.8 ([denoland/deno#33252](https://github.com/denoland/deno/pull/33252)). It downloads the binary built by CI for a given denoland/deno PR via the `gh` CLI and installs it, optionally to a path with `--output`.

- New "Install a build from a pull request" section in `runtime/reference/cli/upgrade.md`.
- Notes the `gh` CLI prerequisite, `#`-prefix tolerance, the `--output`/`--dry-run` interactions, and the `{profile}-{os}-{arch}-deno` artifact-naming scheme.

## Test plan

- [x] `deno task serve` — section renders.

### DIFF
--- a/runtime/reference/cli/upgrade.md
+++ b/runtime/reference/cli/upgrade.md
@@ -97,3 +97,30 @@ the `--canary` build flag for the latest canary build:
 # Upgrade to the latest canary build
 deno upgrade --canary
 ```
+
+## Install a build from a pull request
+
+Starting in Deno 2.8, `deno upgrade pr <number>` downloads the binary built by
+CI for a specific deno PR and installs it. This is handy when you need to
+verify a fix before it has shipped in a release.
+
+```sh
+# Install the binary built by CI for PR #12345
+deno upgrade pr 12345
+
+# Numeric prefix is also accepted
+deno upgrade pr '#12345'
+
+# Write to a path instead of replacing the current binary
+deno upgrade --output ./deno-test pr 12345
+
+# See what would be downloaded without replacing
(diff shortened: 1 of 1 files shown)


fibibot (commented):
The new `deno upgrade pr <number>` section matches denoland/deno#33252: the examples use the accepted `pr` subcommand shape, `--output` and `--dry-run` are parsed before `pr`, and the `gh`/artifact description matches the implementation. No checks are reported on this head, so holding approval until CI is green.

- nit: `last_modified: 2026-03-05` should be bumped for this content addition.
- nit: `Numeric prefix is also accepted` is a little off; this example is accepting a `#` prefix.

fibibot (changes requested):
Thanks for documenting this. The command behavior matches denoland/deno#33252 and the current `cli/tools/upgrade.rs` implementation, but I think this needs one docs-policy fix before merge:

- `runtime/reference/cli/upgrade.md` has `last_modified: 2026-03-05` unchanged while this PR adds a substantive new section. Repo policy asks substantive content changes to bump `last_modified`, so please update it.
- While touching it, please also change the comment above `deno upgrade pr '#12345'` from "Numeric prefix is also accepted" to something like "A # prefix is also accepted". The parser strips `#`; it is not a numeric prefix.\n\nI did not find a factual issue with the `deno upgrade pr` behavior itself.

bartlomieju (approved):

PR #3089, opened 2026-04-29
docs: promote OTEL console exporter to a dedicated section

lunadogbot:
## Summary

The OTEL console exporter ([denoland/deno#32717](https://github.com/denoland/deno/pull/32717)) was previously only described in a tip block inside Quick start. Promotes it to a dedicated "Console exporter" section so it gets a TOC entry and is easier to discover.

- New section in `runtime/fundamentals/open_telemetry.md` after the Quick start.
- Adds a metric line to the example output and a short "when to use" list.
- Cross-links to the existing Configuration section for non-console protocols (http/protobuf, http/json, grpc).
- Leaves the existing tip block in Quick start alone — readers who want the tip-format hint still see it; those scanning the TOC now find a real section too.

## Test plan

- [x] `deno task serve` — TOC shows new "Console exporter" entry, anchor `#console-exporter` resolves, link to `#configuration` works.

### DIFF
--- a/runtime/fundamentals/open_telemetry.md
+++ b/runtime/fundamentals/open_telemetry.md
@@ -92,6 +92,41 @@ You can also create your own metrics, traces, and logs using the
 `npm:@opentelemetry/api` package.
 [Learn more about user defined metrics](#user-metrics).
 
+## Console exporter
+
+For local development and writing instrumentation, set
+`OTEL_EXPORTER_OTLP_PROTOCOL=console` to make Deno write spans, logs, and
+metrics directly to stderr in a human-readable text format. No OTLP collector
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. Line 127 tells production readers to switch to `grpc`, but the same page's Configuration and Limitations sections list only `http/protobuf`, `http/json`, and `console`, and explicitly say `grpc` is not supported. Either remove `grpc` here or update those sections in this PR so the page is internally consistent.

- nit: `last_modified: 2026-03-25` was not bumped for this new section.
- nit: the example-output fence at line 108 should be tagged `console`.

bartlomieju:
The "tip" section we have seems better.

PR #3091, opened 2026-04-29
docs: document `deno eval` CJS/ESM auto-detection

lunadogbot:
## Summary

Documents the new CJS/ESM auto-detection heuristic in `deno eval` shipping in Deno 2.8 ([denoland/deno#32472](https://github.com/denoland/deno/pull/32472)). `deno eval` parses the snippet, treats it as ESM if it contains `import`/`export` declarations, and otherwise runs it as CommonJS — so `require()` works without `--ext=cjs`.

- Replaces the existing "CommonJS support" section in `runtime/reference/cli/eval.md` with "CommonJS and ESM auto-detection".
- Side-by-side example for both detected paths.
- Calls out `--ext=mjs` / `--ext=cjs` as the override when the heuristic gets it wrong.

## Test plan

- [x] `deno task serve` — section renders, examples readable.

### DIFF
--- a/runtime/reference/cli/eval.md
+++ b/runtime/reference/cli/eval.md
@@ -24,13 +24,24 @@ TypeScript works out of the box:
 deno eval "const greeting: string = 'Hello'; console.log(greeting)"
 ```
 
-## CommonJS support
+## CommonJS and ESM auto-detection
 
-CommonJS modules are automatically recognized and supported:
+Starting in Deno 2.8, `deno eval` inspects the snippet for `import` /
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `Otherwise the snippet runs as CommonJS` is no longer true for denoland/deno#32472 as merged. The final change only selects CJS when script syntax also contains CJS-specific patterns (`require`, `module.exports`, `__dirname`, etc.); plain snippets without import/export still default to ESM. Please describe that narrower heuristic, and update the override example accordingly.

- nit: `last_modified: 2025-03-10` should be bumped for this content change.

PR #3092, opened 2026-04-29
docs: document `deno doc npm:` / `deno doc jsr:` (2.8)

lunadogbot:
## Summary

Documents the new ability for `deno doc` to accept `npm:` and `jsr:` specifiers, shipping in Deno 2.8 ([denoland/deno#32435](https://github.com/denoland/deno/pull/32435)).

- New "Documenting an npm or JSR package" subsection in `runtime/reference/cli/doc.md`.
- Examples for both `npm:` and `jsr:` specifiers, version pinning, and combining with `--html` / `--json`.

## Test plan

- [x] `deno task serve` — section renders.

### DIFF
--- a/runtime/reference/cli/doc.md
+++ b/runtime/reference/cli/doc.md
@@ -38,6 +38,22 @@ function add(x: number, y: number): number
   Adds x and y. @param {number} x @param {number} y @returns {number} Sum of x and y
 ```
 
+### Documenting an npm or JSR package
+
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `Starting in Deno 2.8` is too late. On current stable `deno 2.7.14`, `deno doc npm:code-block-writer` and `deno doc --json npm:code-block-writer` both succeed, so this should not present the npm/JSON support as new in 2.8. Please either remove the version claim or narrow it to only the part that actually lands in 2.8.

PR #3096, opened 2026-05-02
docs: document `catalog:` protocol for workspace dependencies (2.8)

lunadogbot:
## Summary

Documents the new `catalog:` protocol shipping in Deno 2.8 ([denoland/deno#32947](https://github.com/denoland/deno/pull/32947)). Catalogs let a workspace root declare shared dependency versions in one place; members reference them with `\"pkg\": \"catalog:\"` or `\"pkg\": \"catalog:<name>\"`.

- Adds a new "Centralized dependency versions with `catalog:`" section to `runtime/fundamentals/workspaces.md`.
- Covers the default catalog, named catalogs (`catalogs`), and catalogs declared in `package.json`.
- Notes the precedence rule when both `deno.json` and `package.json` define catalogs (package.json wins, no merging) and the root-only restriction.

## Test plan

- [x] `deno task serve` — workspaces page renders, new section sits cleanly between "Sharing and managing dependencies" and "Using workspace protocol in package.json".
- [x] `deno fmt` — clean.

### DIFF
--- a/runtime/fundamentals/workspaces.md
+++ b/runtime/fundamentals/workspaces.md
@@ -660,6 +660,113 @@ This approach allows you to:
 3. Test and develop interdependent modules together
 4. Gradually migrate monolithic codebases to modular architecture
 
+## Centralized dependency versions with `catalog:`
+
+When several workspace members depend on the same npm package, keeping their
+versions in sync usually means editing every `package.json` (or `deno.json`)
+whenever you bump a version. The `catalog:` protocol — added in Deno 2.8 and
+compatible with the equivalent feature in pnpm, Bun, and Yarn — lets the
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `package.json` (or `deno.json`) overstates where `catalog:` can be used. denoland/deno#32947 adds `PackageJsonDepValue::Catalog`, and the config schema describes catalogs as being for workspace member `package.json` files; `catalog:` is not an import-map / `deno.json` dependency protocol. Please remove `(or `deno.json`)` or explicitly say only `package.json` dependencies reference catalogs, while root `deno.json` / `package.json` can declare them.

- nit: `last_modified: 2025-11-04` should be bumped for this content addition.

PR #3099, opened 2026-05-05
docs: add module customization hooks reference page

bartlomieju:
## Summary

- Adds a new reference page documenting `module.register()` and `module.registerHooks()` APIs
- Covers resolve/load hooks, hook chaining (LIFO), deregistration, passing data via `initialize()`
- Includes practical examples: virtual modules, custom transpilation, module aliasing, test mocking
- Adds sidebar entry under Reference guides
- Removes outdated "register is a non-functional stub" note from `node:module` compatibility status

These APIs are being implemented in https://github.com/denoland/deno/pull/28985 and allow users to fully customize module resolution and loading, following the Node.js module customization hooks specification.

## Test plan

- [ ] Verify page renders correctly on the docs site
- [ ] Verify sidebar link works
- [ ] Verify node:module compatibility table no longer shows the stub warning

### DIFF
--- a/reference_gen/node_descriptions/module.yaml
+++ b/reference_gen/node_descriptions/module.yaml
@@ -1,3 +1,1 @@
 status: good
-symbols:
-  Module: The `register` method is a non-functional stub.
--- a/reference_gen/node_descriptions/module.yaml
+++ b/reference_gen/node_descriptions/module.yaml
@@ -1,1 +1,3 @@
 status: good
+symbols:
+  Module: The `register` method is a non-functional stub.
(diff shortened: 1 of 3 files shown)


lunadogbot (changes requested):
1. `reference_gen/node_descriptions/module.yaml` is generated content. Removing `symbols:` and the `Module` note here will be overwritten by `deno task generate:reference`; update the upstream source/generator instead, or leave this generated file untouched.
2. CI is failing at `deno fmt --check`, and it reproduces locally in `runtime/reference/module_hooks.md` starting at `resolve(specifier, context, nextResolve)`. Run `deno fmt runtime/reference/module_hooks.md runtime/_data.ts reference_gen/node_descriptions/module.yaml` and commit the result.

PR #3101, opened 2026-05-07
docs: add structured clone and transferable API reference

bartlomieju:
## Summary

- Adds tables documenting which Web API types are serializable (support `structuredClone()` / `postMessage()`) and which are transferable
- Updated both the API reference page (`api/web/index.md`) and the runtime reference page (`runtime/reference/web_platform_apis.md`)
- Motivated by the addition of Blob/File structured clone support in denoland/deno#33827

### Serializable types documented
Primitives, Array, Object, Map, Set, Date, RegExp, ArrayBuffer, TypedArray, DataView, Error types, Blob, File, DOMException, CryptoKey

### Transferable types documented
ArrayBuffer, MessagePort, ReadableStream, WritableStream, TransformStream

## Test plan
- [ ] Verify tables render correctly on the docs site
- [ ] Verify links resolve properly

### DIFF
--- a/api/web/index.md
+++ b/api/web/index.md
@@ -72,6 +72,54 @@ Deno supports a comprehensive set of web standard APIs:
 - **[Timers](/api/web/~/setTimeout)** - setTimeout, setInterval, and
   setImmediate
 
+### Structured Clone & Transferable Objects
+
+Deno supports [`structuredClone()`](/api/web/~/structuredClone) and
+[`postMessage()`](/api/web/~/Worker) for cloning and transferring objects across
+contexts (e.g. between the main thread and Web Workers).
+
+#### Serializable types
+
+These types can be cloned with `structuredClone()` and sent via `postMessage()`:
+
+| Type | Notes |
+| --- | --- |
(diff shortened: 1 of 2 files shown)


lunadogbot (changes requested):
Two blockers:

1. `deno fmt --check` fails on both new tables — column alignment. Running `deno fmt api/web/index.md runtime/reference/web_platform_apis.md` fixes it.

2. The "Clone a Blob" sample doesn't run on current stable. denoland/deno#33827 merged 2026-05-07, but the most recent release is 2.7.14 (2026-04-28), so on stable `structuredClone(new Blob([...]))` returns `{}` and the example throws `TypeError: cloned.text is not a function`. Same for the `File` row. Either gate this on the next release (note "Requires Deno 2.8+" near Blob/File) or hold the PR until 2.8 ships.

- nit: `runtime/reference/web_platform_apis.md` has `last_modified: 2025-05-09` — bump it on substantive additions.
- nit: the transfer example uses a `worker` symbol that isn't defined in the snippet. `const ch = new MessageChannel(); ch.port1.postMessage(buffer, [buffer]);` is self-contained and demonstrates the same point.

PR #3102, opened 2026-05-07
docs: document Deno.test() timeout option

bartlomieju:
## Summary

- Add a "Timeouts" section to the testing fundamentals page
  documenting the per-test `timeout` option for `Deno.test()`
- Covers async hangs, sync hot loops, and the `timeout: 0`
  disable behavior

Corresponds to denoland/deno#33815.

### DIFF
--- a/runtime/fundamentals/testing.md
+++ b/runtime/fundamentals/testing.md
@@ -118,6 +118,30 @@ Deno.test("database operations", async (t) => {
 });
 ```
 
+## Timeouts
+
+You can set a maximum duration for individual tests using the `timeout` option.
+If a test exceeds its deadline it is marked as failed. Both asynchronous hangs
+(a promise that never resolves) and synchronous hot loops (`while (true) {}`)
+are caught.
+
+```ts
+Deno.test({
+  name: "completes within deadline",
+  timeout: 5000, // 5 seconds
+  async fn() {
+    const response = await fetch("https://example.com");
+    await response.body?.cancel();
+  },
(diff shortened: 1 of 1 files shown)


fibibot (commented):
Wording on the `timeout` option, `0`/unset semantics, and "next test still runs" match the upstream PR (denoland/deno#33815) — the spec tests `async_hang.out` and `sync_hot_loop.out` assert that exact behavior, and the new docstring on `Deno.TestDefinition.timeout` in `lib.deno.ns.d.ts` says the same thing.

One claim I'd like to confirm before approving: "Sanitizers are skipped for the timed-out test since a forcibly terminated test will naturally have leaked resources or operations." That's not in the upstream TS docstring, and I don't see explicit `TimedOut`-aware branching around the sanitizer pass in `cli/tools/test/mod.rs`. Can you point at where this happens, or soften the claim if it's an inference about what users will observe rather than a guaranteed behavior?

Holding approval until `lint and link check` is green (currently CANCELLED on this SHA — needs a re-run).

fibibot (approved):
Verified the sanitizer-skip claim against upstream spec `tests/specs/test/timeout/async_hang.ts` — the pending `setTimeout(resolve, 1_000_000)` would surface as an op-sanitizer failure if sanitizers ran on the timed-out test, but `async_hang.out` only shows `error: Test timed out after 100ms.` with no sanitizer output. The doc statement matches observed behavior. CI green; approving.

PR #3107, opened 2026-05-13
Rewrite FFI structs section with a runnable example

lunadogbot:
The "Working with structs" section in `runtime/fundamentals/ffi.md` had an
example that did not run: it defined the struct as `{ fields: { x: "f64", y:
"f64" } }` (the supported-types table on the same page documents `{ struct:
[...] }` as an array of field types), called `new Deno.UnsafePointer(...)` as
a constructor, and built struct values by passing `BigInt(Float64Array.of(1.0)
.buffer)` — which throws `TypeError: Cannot convert [object ArrayBuffer] to a
BigInt`. A reader copying it would hit that error before getting anywhere.

This PR replaces the section with a complete walkthrough: a small C `Point`
library, per-platform build commands (Linux/macOS/Windows), and a Deno script
that calls `distance` and `midpoint` using `{ struct: ["f64", "f64"] }`,
passing values as a `Uint8Array` view of a `Float64Array` and decoding the
returned struct back through a `Float64Array`. The observed `console.log`
output is shown verbatim. A short "things to keep in mind" list calls out
padding, positional fields, and how returned structs come back as bytes.

Verified by building `libpoint.so` with `cc 13.3.0` and running the exact
script from the page inside `denoland/deno:latest` (2.7.14 — no `:canary`
tag is currently published) with `--allow-ffi`. Output matched: `distance =
5`, `midpoint = { x: 2.5, y: 4 }`. `deno fmt` was run on the file. The
section's only change is to itself; no other sections were touched.


bartlomieju (commented):
Nice work, please remove em dashes from prose

lunadogbot (commented):
I verified the FFI struct example locally on Deno 2.7.14: `cc -shared -fPIC -O2 -o libpoint.so point.c` built the shared library, and `deno run --allow-ffi point.ts` printed `distance = 5` and `midpoint = { x: 2.5, y: 4 }` as documented.

Holding off on approval because the earlier maintainer feedback is still unresolved: the added prose still contains em dashes in this section.

lunadogbot (commented):
I tested the FFI sample locally on Deno 2.7.14 with cc 13.3.0. The C library builds and the exact TypeScript sample prints the documented output:

```console
distance = 5
midpoint = { x: 2.5, y: 4 }
```

Holding off approval because the existing maintainer comment still applies: the new prose still contains em dashes.

PR #3109, opened 2026-05-13
Replace contrived hashbang example with Deno.args greeter

lunadogbot:
The "Creating an example script" section in `examples/tutorials/hashbang.md`
opened with a script that printed `Deno.env.get("DENO_INSTALL")`. That env
var is only populated when Deno is installed via the official install
script — for readers on Homebrew, apt, the Docker images, or any other
install path, the very first example in the tutorial prints `undefined`,
which is a confusing first impression of "executable scripts."

This PR rewrites the section to use a `Deno.args`-based greeter instead.
The new example needs no `--allow-*` flags (a useful teaching beat about
Deno's default-deny permissions), always produces reproducible output, and
demonstrates the shape almost every CLI tool starts with — argument
parsing with a default. The `-S` explanation is also tightened: the old
text said `-S` "indicates that the following argument should be passed to
the env command," which doesn't really say what `-S` does; the new text
explains that `-S` splits the single string into separate arguments so
`env` doesn't try to look up a program literally named `deno run`. The
`console` block now shows both invocations and their output.

Verified by running the new example against `denoland/deno:latest`
(Deno 2.7.14 — the `denoland/deno:canary` tag isn't currently published on
Docker Hub) both as `deno run hashbang.ts` and via the hashbang directly
after `chmod +x`. With no arg it prints `Hello, world!`; with `Ada` it
prints `Hello, Ada!`. `last_modified` bumped to today.

@bartlomieju ready for review when you have a moment.

### DIFF
--- a/examples/tutorials/hashbang.md
+++ b/examples/tutorials/hashbang.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2025-03-10
+last_modified: 2026-05-21
 title: "Executable scripts"
 description: "Guide to creating executable scripts with Deno. Learn about hashbangs, file permissions, cross-platform compatibility, and how to create command-line tools that can run directly from the terminal."
 url: /examples/hashbang_tutorial/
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):

examples/tutorials/hashbang.md
```diff
@@ -28,40 +28,49 @@ Linux (WSL) or use a Unix-like shell like
...
-flag splits the command into arguments and indicates that the following argument
-(`deno run --allow-env`) should be passed to the env command.
+The first line is the hashbang. It tells the system to run the file with
+`/usr/bin/env`, which then locates `deno` on your `PATH` and invokes `deno run`
+on the script. The `-S` flag splits `deno run` into separate arguments so `env`
+treats them correctly; without `-S` the whole string would be passed as a single
+argument name and `env` would fail to find a program called `deno run`. There
+are no `--allow-*` flags here because the script does not read the filesystem,
+network, or environment — Deno only asks for permissions when the script
+actually needs them.
```
bartlomieju: Good, but make this section a bit shorter and remove em dashes

PR #3117, opened 2026-05-14
fix: swap mismatched console.log calls in importing_bytes example

### LINKED ISSUE #3044: Feedback: /examples/importing_bytes/ - Positive
## Documentation Feedback

**Path:** [/examples/importing_bytes/](https://docs.deno.com/examples/importing_bytes/)
**Helpful:** Yes ✅

**Feedback:**
> It seems that you have a tiny mistake in the code:
you need to change this line:  console.log(text);
to be:  console.log(bytes);

---

lunadogbot:
## Summary

The `/examples/importing_bytes/` snippet logs the wrong variable in both halves of the example. As pointed out by the reader in #3044:

- Line 20 logs `text` immediately after `import bytes from "./image.png"` — `text` is not defined yet (TDZ).
- Line 26 logs `bytes` after `const text = await import(...)` — should log `text`.

Swap the two `console.log` arguments so each log matches its preceding import. Variable names are left as-is to keep the diff minimal, per the reporter's request.

## Test plan
- [x] Re-read the snippet end-to-end and confirmed both `console.log` calls now reference variables that exist at that point in the program.

Closes #3044
Closes bartlomieju/orchid-inbox#61

cc @bartlomieju

### DIFF
--- a/examples/scripts/importing_bytes.ts
+++ b/examples/scripts/importing_bytes.ts
@@ -17,13 +17,13 @@
 // Binary files can be imported in JS and TS modules. When doing so, you need to
 // specify the `type: "bytes"` import attribute.
 import bytes from "./image.png" with { type: "bytes" };
-console.log(text);
+console.log(bytes);
 
 // Dynamic imports are also supported.
 const text = await import("./image.png", {
   with: { type: "bytes" },
 });
-console.log(bytes);
+console.log(text);
 
 /* File: ./image.png
 89 50 4e 47 0d 0a 1a 0a 00 00 00 0d 49 48 44 52


fibibot (commented):
Diff swap is correct. Pre-fix, line 20 `console.log(text)` hits the TDZ (the `const text` is on line 23), and line 26 `console.log(bytes)` logs the static-import value under a comment that introduces dynamic imports. After the swap, each `console.log` references the binding immediately above it, which is the intent. The diff is also intentionally minimal — no rename of `text` (which still holds the dynamic-import namespace object, not bytes), matching the reporter's request in #3044.

Holding approval until `lint and link check` finishes. The currently-failing run is the same `/styleguide/typography.mdx` "Cannot resolve module 'lume/jsx-runtime'" build error affecting other open PRs (e.g. #3107), not anything in `examples/scripts/importing_bytes.ts`.

bartlomieju (approved):

PR #3122, opened 2026-05-14
Fix denoland/docs#2935: multi-stage Docker build doesn't ship deps to the runtime stage

fibibot:
Closes bartlomieju/orchid-inbox#65


bartlomieju (changes requested):
_config.ts and _plugins changes appear to be unnecessary, revert them

bartlomieju (approved):

PR #3127, opened 2026-05-14
expand mysql2 querying section with parameterized queries

lunadogbot:
Restructures the "Querying MySQL" section of `examples/tutorials/mysql2.md`. The
old section had two problems: the first SELECT did not destructure the
`[rows, fields]` tuple, so the printed output started with an outer `[` that
wrapped both rows and fields and was then truncated mid-array; and the
single-row example used `WHERE \`name\` = 'Deno'` — a string-interpolation
pattern that teaches readers exactly the antipattern that ships SQL injection
bugs.

The new section keeps the same flow but uses `[rows]` destructuring in the
first example (with a closed, accurate output block) and replaces the
single-row example with a `connection.execute()` call that uses a `?`
placeholder, plus a short prose note explaining why parameterized queries are
mandatory for any user-supplied value. The "Common pitfalls" / "what about
INSERT/UPDATE/DELETE" angle is covered in one paragraph rather than a wall of
prose.

Verified end-to-end against a MySQL 8 container using the same version pin the
page imports (`npm:mysql2@^2.3.3/promise`) under `denoland/deno:latest`
(2.7.14). Both code blocks were run as written and the output blocks reflect
what the driver actually returns.

### DIFF
--- a/examples/tutorials/mysql2.md
+++ b/examples/tutorials/mysql2.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2025-03-10
+last_modified: 2026-05-14
 title: "How to use MySQL2 with Deno"
 description: "Step-by-step guide to using MySQL2 with Deno. Learn how to set up database connections, execute queries, handle transactions, and build data-driven applications using MySQL's Node.js driver."
 url: /examples/mysql2_tutorial/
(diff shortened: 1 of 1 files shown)


fibibot (approved):
`mysql2/promise`'s `.query()` resolves to `[rows, fields]`, so the old single-binding version was printing the tuple and the output's closing `]` got eaten — `[rows]` destructure fixes it. `.execute()` with `?` placeholders is the right swap; mysql2 does send these as real `COM_STMT_PREPARE`/`COM_STMT_EXECUTE` and caches the prepare on repeats, so the prose is accurate.

`last_modified` is bumped; the new `"Abels lizard"` output matches the `INSERT` two sections up (no apostrophe in the row, no stray `"` like the old output had).

bartlomieju (approved):

PR #3143, opened 2026-05-20
docs: document Network tab debugging for fetch/WebSocket (2.8)

bartlomieju:
Deno 2.8 wires `fetch()`, `node:http` / `node:https` client requests, and
`WebSocket` into the Chrome DevTools Network tab via the Network CDP
domain, but the debugging guide didn't mention any of this. This adds a
new "Inspecting network traffic" section to `runtime/fundamentals/debugging.md`
that explains how to attach DevTools with `--inspect-wait` and what the
Network tab now shows, with a small `fetch()` example and a note that
the instrumentation has no overhead when no debugger is attached.

Image references are included as placeholders to be filled in before
merging.


bartlomieju (commented):
I think it only works when `Chrome DevTools for Node.js` is used:

Image

PR #3169, opened 2026-05-27
ci: unpin Deno version in lint workflow

bartlomieju:
The lint workflow pinned Deno to v2.x via setup-deno, which is no longer
necessary now that v2 is the default stable channel. Dropping the pin
lets CI track the latest stable Deno release automatically and keeps the
workflow consistent with the other workflows in this repo, which already
omit deno-version.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,7 +23,6 @@ jobs:
         uses: denoland/setup-deno@v2
         with:
           cache: true
-          deno-version: v2.x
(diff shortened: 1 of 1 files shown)


fibibot (commented):
`denoland/setup-deno@v2` already defaults `deno-version` to `2.x`, so removing the explicit input keeps the lint workflow on the same stable channel while avoiding a redundant setting. Holding approval until Deploy is green.

PR #3172, opened 2026-05-27
docs: tidy up runtime landing page

bartlomieju:
This cleans up a handful of small inconsistencies on the runtime landing
page (`runtime/index.md`) without touching anything else.

The "secure by default" bullet under "Why Deno?" was missing the word
"to" - it now reads "give full access to every script they run". The
"Deno is fast and reliable" bullet was unlinked and read as filler next
to the other linked claims, so I removed it rather than inventing a link
that does not exist; if we later want to point at a benchmarks or
performance page, it can come back with a real target.

The "First steps" list previously jumped straight from the landing page
to First project, Setup, and CLI, skipping Installation even though
Installation is the very next entry in the runtime sidebar
(`runtime/_data.ts`). I added "Installing Deno" as the first item so the
list mirrors the sidebar order. While there, the CLI link was missing
its trailing slash (`/runtime/getting_started/command_line_interface`
vs. every other internal link in the file), so I added it.

The macOS and Linux tabs in "Quick install" still run the same
`install.sh` one-liner, which is correct - the tabs exist because users
expect to pick their OS, not because the command differs. To make the
Linux tab pull its weight I added a short note that the installer needs
`unzip` (or `7z` on Alpine) available on the system, which is a real
gotcha on minimal Linux images and not something macOS users hit.

Ran `deno fmt runtime/index.md` after the edits; no other files
touched.

### DIFF
--- a/js/tabs.ts
+++ b/js/tabs.ts
@@ -79,6 +79,15 @@ for (const tabGroup of document.querySelectorAll<HTMLElement>(".deno-tabs")) {
   const storedTabId = localStorage.getItem(`deno-tab-${groupId}`);
   if (storedTabId) {
     document.dispatchEvent(new GroupSelectEvent(groupId, storedTabId));
+  } else if (groupId === "operating-systems") {
+    const ua = navigator.userAgent;
+    let detected: string | null = null;
+    if (/Mac/i.test(ua)) detected = "mac";
+    else if (/Win/i.test(ua)) detected = "windows";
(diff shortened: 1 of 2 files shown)


fibibot (approved):
The changed internal links resolve in the repo, and `install.sh` currently exits unless `unzip` or `7z` is available, so the Linux tab note matches the installer. No issues found.

PR #3176, opened 2026-05-27
docs: tidy CLI overview flag sections and ordering warning

bartlomieju:
This PR cleans up a few small issues on the Command line interface
getting-started page (`runtime/getting_started/command_line_interface.md`)
that made the flag reference sections confusing without changing the
underlying information.

The inline flag listings for integrity (lock files), cache and
compilation flags, and "Other runtime flags" were previously fenced
shell-style code blocks that visually clashed with the existing
type-checking table on the same page. They are now small markdown
tables that match that style, which makes them easier to scan and
keeps the page visually consistent. No flags were added or removed in
the conversion; only formatting changed, with minor wording
normalization on a couple of descriptions.

The `### Runtime flags` heading previously had no real content under
it — just a one-sentence intro followed immediately by the next `###`
heading. Rather than fabricate a flag list (the actual runtime-only
flags such as `--prompt`, `--seed`, `--location`, `--inspect*`, and
`--v8-flags` already live in the "Other runtime flags" subsection
further down), the heading is kept as a short orienting paragraph
that explains the subsections below it are the runtime-flag
categories. This avoids a duplicated flag list while still giving the
heading a purpose.

The "Argument and flag ordering" section opened with an italicized
"Note that…" sentence about runtime flags being consumed as script
arguments when placed after the script name. This is a genuinely
common footgun, so it has been promoted into a `:::caution`
admonition — that is the callout style documented in the repo's own
`styleguide/writing.md` and used elsewhere (for example
`deploy/kv/index.md`). The example block immediately below it is
unchanged.

The page previously had no reference to `--unstable-*` flags despite
unstable feature flags having a dedicated sidebar entry. A short
pointer to `/runtime/reference/cli/unstable_flags/` has been added at
the end of the "Other runtime flags" section so readers can discover
the page from the CLI overview.

Finally, the Hot Module Replacement section now ends with a brief
clarifying sentence noting that `--watch-hmr` is a server-side
runtime feature for Deno processes and is not a browser-style HMR
mechanism like Vite's. This avoids confusion for readers coming from
a frontend tooling background who might otherwise expect updates to
be pushed into a running browser page automatically.

The file was run through `deno fmt` after editing. No other files
were touched.

### DIFF
--- a/runtime/getting_started/command_line_interface.md
+++ b/runtime/getting_started/command_line_interface.md
@@ -16,8 +16,8 @@ standalone binaries, and a lot more. Each subcommand (`run`, `test`, `fmt`,
 `deno <subcommand> --help` to see them.
 
 For the complete list of subcommands and flags, see the
-[CLI reference](/runtime/reference/cli/). The rest of this page walks through
-the most useful flags in more detail.
+[CLI reference](/runtime/reference/cli/). This page covers the patterns you'll
+hit early on: how to run code, how to pass arguments, and how to use watch mode.
 
 ## Running scripts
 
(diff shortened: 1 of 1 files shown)


fibibot (changes requested):
1. `last_modified: 2025-11-05` is unchanged, but this PR adds new user-facing CLI guidance for flag ordering, HMR, and unstable flags. Please bump the date so the page freshness metadata reflects the content change.

PR #3198, opened 2026-06-10
docs: layout & spacing overhaul

bartlomieju:
Overhauls the docs layout and spacing so pages feel less cramped and more
readable. Everything is driven from a single `--layout-max-width` CSS variable
so the header, content, deploy SubNav, and footer all stay aligned.

## Layout & width
- Widen the content layout to **90rem** and rebalance the three-column grid
  (`minmax(13rem,14rem)` sidebar / `minmax(0,1fr)` content / `16rem` TOC).
- Tie the deploy **SubNav** and the **footer** widths to `--layout-max-width`
  so they line up with the header and content.

## Left sidebar
- More vertical breathing room on nav items (`py-1` → `py-2`).
- Denser nested items with a single guide line (no per-item border) and an
  open-parent highlight; narrow the sidebar now that labels are shorter.
- Tidy labels, slim the **Contributing** section to a few flat items, and
  rename the *module hooks* page to *loader hooks*.
- Add a test that imports the sidebar data and walks it, capping every
  sidebar (runtime, deploy, deploy classic, subhosting, sandbox) at its
  current longest label so new labels can't blow out the nav.

## Right sidebar (table of contents)
- Denser text, wider column, hidden scrollbar, a **"This page"** header, and a
  green active accent. Cap nesting at 2 levels (3rd/4th-level headings added
  clutter without aiding navigation).
- Fix scroll-spy so it no longer drops the active item (track only headings
  that have TOC links and mark the last one passed).
- Fix the TOC disappearing on short pages (e.g. `/deploy/`): it now uses
  `max-height` instead of a fixed height, so it collapses to its content and
  keeps the slack `position: sticky` needs to stay pinned.

## Content polish
- Condense code-block padding.
- **Tables:** move off the clashing GitHub canvas/border colors onto Deno
  design tokens (subtle borders, rounded container, header background + zebra),
  and cap cell width so long cells wrap instead of forcing a wide scroll. Fix
  the fmt-options table columns (Option column was hogging ~half the width).
- **Feedback box:** replace the heavy filled panel with a light top-border
  separator and put the question + Yes/No/Edit on a single compact row.

## SubNav
- Mark the active SubNav item with a persistent underline so the selection is
  clear against the green bar.


sidebar_test.ts
```diff
@@ -0,0 +1,35 @@
...
+// shorten another) — only raise this number with a deliberate design change.
+const MAX_LABEL_LENGTH = 22;
+
+// We read the `title:` string literals out of runtime/_data.ts directly rather
+// than importing it: importing pulls in types.ts, whose `JSX.Element` type does
+// not resolve under `deno test`'s standalone type-check. This covers every
+// human-written label (the ones at risk of being too long); the generated
+// standard-library entries use a variable and are short JSR module names.
+Deno.test("Runtime sidebar labels stay within the max length", async () => {
+  const source = await Deno.readTextFile("./runtime/_data.ts");
```
bartlomieju: Import this file directly in this test file and just walk JS objects, don't rely on regexes, it's too brittle

bartlomieju: Also apply the same check to the Deploy sidebars

PR #3209, opened 2026-06-11
docs: SEO batch — canonicals, titles, breadcrumbs, prev/next

bartlomieju:
A content review of the runtime docs found several site-wide SEO gaps: no
rel=canonical anywhere (risky while we restructure URLs behind 301s), page
titles with no site name (a SERP entry for "deno test" read as just "deno
test"), no structured data, and no way to read a section linearly. This
closes them at the layout level: every page now emits a canonical URL
(except 404), titles get a "| Deno Docs" suffix (og:title stays bare on
purpose), and a BreadcrumbList JSON-LD trail (Deno Docs → section → page,
using each section's sectionTitle) gives search engines the hierarchy even
though we don't render visible breadcrumbs yet — that UI is deferred to the
upcoming styling overhaul.

Doc pages also gain prev/next links derived from the section sidebar's
reading order, reusing Navigation's own getSectionData so pagination can
never disagree with the sidebar; API reference and lint-rule pages are
excluded. The copy-page menu picks up an llms.txt entry alongside "View as
Markdown" and "Open in Claude" so the machine-readable index is
discoverable from every page. Verified in the rendered build across
runtime, deploy (subsection sidebars), examples, and 404 pages.


_includes/layout.tsx
```diff
@@ -8,6 +8,38 @@ export default function Layout(data: Lume.Data) {
...
+  const sectionRoot = section ? `/${section}/` : null;
+  const sectionTitle = sectionRoot
+    ? (data.search.data(sectionRoot)?.sectionTitle ??
+      section.charAt(0).toUpperCase() + section.slice(1))
+    : null;
+  const breadcrumbJsonLd = pageTitle && !is404
+    ? {
+      "@context": "https://schema.org",
+      "@type": "BreadcrumbList",
+      itemListElement: [
```
bartlomieju: What is the purpose of this? What tools use it?

bartlomieju: This is schema.org BreadcrumbList structured data (JSON-LD). Search engines consume it: Google and Bing use it to render a breadcrumb trail in the search result (instead of the raw URL) and to understand the site hierarchy — see https://developers.google.com/search/docs/appearance/structured-data/breadcrumb. It's metadata only; we don't render a visible breadcrumb UI yet (deferred to the styling overhaul), but this gets the SERP benefit now. Expanded the code comment to say this: 6b372385.

PR #3214, opened 2026-06-11
docs: add package management tutorials

bartlomieju:
Package management was the largest task-shaped gap left in the examples
catalog: workflows like private registries, lockfile handling, lifecycle
scripts, and workspaces lived only as reference prose, with no findable
task pages. This adds eight tutorials under "Modules and package
management": adding/removing dependencies (incl. aliasing and version
pinning), locking dependencies with deno.lock (incl. frozen-lockfile CI
errors shown verbatim), private npm registries (.npmrc, registry overrides,
Azure Artifacts and Artifactory setups, mTLS), npm lifecycle scripts and
approve-scripts, local and unpublished packages (the links field, HTTPS
imports, and an honest section on what specifier types are not supported),
workspaces/monorepos (incl. catalog:), running Deno in GitHub Actions, and
using Deno in an existing Node.js project.

Command outputs shown in the tutorials were captured from real runs on
Deno 2.8 rather than written from memory, and each page links to the
deeper runtime documentation it summarizes.


examples/tutorials/add_remove_dependencies.md
```diff
@@ -0,0 +1,105 @@
+---
+title: "Add and remove dependencies"
+description: "Manage project dependencies with deno add and deno remove: npm and JSR packages, version pinning, dev dependencies, and aliasing packages in your import map."
+url: /examples/add_remove_dependencies_tutorial/
+---
+
+Deno manages dependencies through your project's `deno.json` import map, or
+through `package.json` when you have one. The `deno add` and `deno remove`
```
bartlomieju, line 8: `deno install` also does that

examples/tutorials/add_remove_dependencies.md
```diff
@@ -0,0 +1,105 @@
...
+commands edit those files for you.
+
+## Adding packages
+
+Unprefixed package names are treated as npm packages. Use the `jsr:` prefix for
```
bartlomieju, line 14: mention `deno install` can be used for that purpose too

examples/tutorials/add_remove_dependencies.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+:::note
+
+Without a `package.json`, the `--dev` flag has no effect: `deno.json` has a
+single `imports` map, so the package is added as a regular import.
+
+:::
+
+## Aliasing a package
```
bartlomieju, line 72: What about `overrides` in `package.json`?

examples/tutorials/dependency_lockfile.md
```diff
@@ -0,0 +1,126 @@
...
+      "integrity": "sha512-7NzBL0rN6fMUW+f7A6Io4h40qQlG+xGmtMxfbnH/K7TAtt8JQWVQK+6g0UXKMeVJoyV5EkkNsErQ8pVD3bLHbA=="
+    }
+  },
+  "workspace": {
+    "dependencies": ["npm:chalk@^5.6.2"]
+  }
+}
+```
+
+Each semver range from your config is mapped to one exact version, and each
```
bartlomieju, line 34: Mention `deno ci` in this tutorial

examples/tutorials/deno_github_actions.md
```diff
@@ -0,0 +1,103 @@
...
+      - run: deno lint
+      - run: deno test --allow-all
+```
+
+`deno-version` accepts a semver range like `v2.x`, an exact version, or
+`canary`.
+
+## Installing with a frozen lockfile
+
+`deno ci` (Deno 2.8+) mirrors `npm ci`: it is roughly equivalent to deleting
```
bartlomieju: We can skip `deno 2.8+`

examples/tutorials/migrate_node_project.md
```diff
@@ -0,0 +1,97 @@
...
+`deno run --allow-net --allow-env main.js`, or start permissive with
+`deno run -A main.js` and tighten later.
+
+:::
+
+Two compatibility rules cover almost everything else:
+
+- **Node built-ins need the `node:` prefix**: write `import os from "node:os"`
+  instead of `import os from "os"`. Node itself supports this prefix, so the
+  change is backwards-compatible.
```
bartlomieju: We got a PR that changes that scheduled for 2.9 we should take care of this one

examples/tutorials/migrate_node_project.md
```diff
@@ -0,0 +1,97 @@
...
+- **CommonJS and ES modules both work**, using the same rules as Node: `.cjs`
```
bartlomieju: MEntion the caveat that by default Deno uses ESM instead of CJS

examples/tutorials/npm_lifecycle_scripts.md
```diff
@@ -0,0 +1,96 @@
...
+## Allow scripts at install time
+
+Pass `--allow-scripts` with the packages you trust:
+
+```sh
+deno install --allow-scripts=npm:better-sqlite3
+```
+
+This installs all dependencies and runs lifecycle scripts only for
+`better-sqlite3` — other packages' scripts stay blocked.
```
bartlomieju: This should be a minoir stuff, `deno approve-scripts` would be the preferred way

examples/tutorials/npm_lifecycle_scripts.md
```diff
@@ -0,0 +1,96 @@
...
+## Install a package that needs build scripts
+
+With a `node_modules` directory enabled, install a package that declares
+lifecycle scripts:
+
+```json title="deno.json"
+{
+  "nodeModulesDir": "auto"
+}
+```
```
bartlomieju: This just complicates stuff :( we should find a way to explain the feature without this one

examples/tutorials/private_registries.md
```diff
@@ -0,0 +1,103 @@
...
+separated by semicolons:
+
+```sh
+DENO_AUTH_TOKENS=a1b2c3d4e5f6@deno.land;username:password@example.com:8080
+```
+
+Deno sends the matching `Authorization` header when fetching modules from those
+hosts.
+
+## Azure Artifacts
```
bartlomieju, line 87: What about github package registry, verdaccio, etc?

examples/tutorials/workspaces_monorepo.md
```diff
@@ -0,0 +1,137 @@
...
+
+The root `deno.json` lists the member directories:
+
+```json title="deno.json"
+{
+  "workspace": ["./utils", "./app"]
+}
+```
+
+Each member gets its own `deno.json` with a `name`, a `version`, and an
```
bartlomieju, line 22: Mention that it can be `package.json` too for workspace member?

examples/tutorials/workspaces_monorepo.md
bartlomieju, line 1: Add examples of using npm workspace and yarn workspace directly. And pnpm workspace with caveat of converting `pnpm-workspace.yml` to `deno.json`. We got a PR that will help with that scheduled for 2.9

bartlomieju:
Addressed all review comments:

- **add_remove_dependencies**: `deno install` now mentioned both in the intro and as the equivalent of `deno add` (verified it writes the same imports entry); new **Overriding transitive dependencies** section showing `overrides` in package.json (verified: forces `ms@2.1.2` into the lockfile).
- **dependency_lockfile**: `deno ci` now appears early (commit-to-git paragraph) in addition to the CI section; dropped the version qualifier.
- **deno_github_actions**: dropped the `Deno 2.8+` qualifier and the older-versions fallback sentence.
- **migrate_node_project**: softened the `node:` prefix wording (easy to revise when the 2.9 change lands) and added a note that Deno assumes ESM by default — a `.js` file is only CJS with explicit `"type": "commonjs"`, unlike Node's implicit default.
- **npm_lifecycle_scripts**: restructured — the example now lives in a package.json project so `node_modules` is automatic and the config preamble is gone (the nodeModulesDir nuance moved to a note); `deno approve-scripts` is now the primary flow, with `--allow-scripts` demoted to a one-line alternative.
- **private_registries**: added **GitHub Packages** and **Verdaccio** sections. Verified empirically that Deno honors `.npmrc` `_authToken` (pointed a scope at a local server and confirmed the `Bearer` header is sent) since the runtime docs only show `_auth`.
- **workspaces_monorepo**: members-can-be-package.json mentioned at the member definition; new **Using existing npm, yarn, and pnpm workspaces** section with a verified `workspaces`-field example (member-to-member import via dependencies), and a note that `pnpm-workspace.yaml` needs its list moved into deno.json/package.json.

All examples with terminal output were re-verified against real runs; tests pass.

PR #3215, opened 2026-06-11
docs: add HTTP server examples

bartlomieju:
Fills the remaining task-shaped HTTP gaps in the examples catalog
(follow-up to #3208/#3210/#3211/#3214): serving over TLS with a local
certificate, server-sent events with proper headers and client
disconnection handling, a streaming reverse proxy built on fetch, hot
reload with --watch, scaling across CPU cores with deno serve --parallel
(including the per-worker-state caveat), fetching over Unix domain sockets
with Deno.createHttpClient (with a Docker-socket example), and serving
Node.js streams from both node:http and Deno.serve. The existing streaming
example also gains the async-iterator pattern via ReadableStream.from
rather than adding a near-duplicate page.

Every server was verified with real requests; outputs in the comments are
captured, not invented. One pre-existing example picked up a small type fix
(setInterval's return type under mixed Deno/Node typings) along the way.

### DIFF
--- a/examples/scripts/fetch_unix_socket.ts
+++ b/examples/scripts/fetch_unix_socket.ts
@@ -1,0 +1,40 @@
+/**
+ * @title Fetch over a Unix socket
+ * @difficulty intermediate
+ * @tags cli
+ * @run -N -R -W <url>
+ * @resource {https://docs.deno.com/api/deno/~/Deno.createHttpClient} Doc: Deno.createHttpClient
+ * @resource {https://docs.deno.com/api/deno/~/Deno.serve} Doc: Deno.serve
+ * @group Network
+ *
+ * Local services like the Docker daemon expose HTTP over a Unix domain
+ * socket instead of a TCP port. This example serves HTTP on a socket file
+ * and fetches from it.
+ */
+
+// Deno.serve listens on a Unix socket when given a path instead of a port.
+const server = Deno.serve(
+  { path: "/tmp/example.sock" },
+  () => new Response("Hello over a Unix socket"),
+);
+
+// On the client side, create an HTTP client that connects through the
+// socket, and pass it to fetch. The URL's hostname is ignored; only the
+// path and headers matter to the server.
+const client = Deno.createHttpClient({
+  proxy: { transport: "unix", path: "/tmp/example.sock" },
+});
+
+const response = await fetch("http://localhost/", { client });
+console.log(await response.text()); // Hello over a Unix socket
+
+client.close();
+await server.shutdown();
+
+// The same client works for talking to existing services. For example,
+// with -R -W -N permissions and Docker running, this lists containers:
+//
+//   const docker = Deno.createHttpClient({
+//     proxy: { transport: "unix", path: "/var/run/docker.sock" },
+//   });
+//   await fetch("http://localhost/containers/json", { client: docker });
--- a/examples/scripts/fetch_unix_socket.ts
+++ b/examples/scripts/fetch_unix_socket.ts
@@ -31,6 +31,24 @@ console.log(await response.text()); // Hello over a Unix socket
 client.close();
 await server.shutdown();
 
+// The Node.js API supports Unix sockets too: a node:http server listens on
+// a path, and the client passes socketPath in the request options.
+import { createServer, request } from "node:http";
+
+const nodeServer = createServer((_req, res) => {
+  res.end("Hello from node:http");
+});
+nodeServer.listen("/tmp/example-node.sock", () => {
+  const req = request(
+    { socketPath: "/tmp/example-node.sock", path: "/" },
+    (res) => {
+      res.on("data", (chunk) => console.log(String(chunk))); // Hello from node:http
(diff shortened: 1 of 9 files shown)


examples/scripts/fetch_unix_socket.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @resource {https://docs.deno.com/api/deno/~/Deno.createHttpClient} Doc: Deno.createHttpClient
+ * @resource {https://docs.deno.com/api/deno/~/Deno.serve} Doc: Deno.serve
+ * @group Network
+ *
+ * Local services like the Docker daemon expose HTTP over a Unix domain
+ * socket instead of a TCP port. This example serves HTTP on a socket file
+ * and fetches from it.
+ */
+
+// Deno.serve listens on a Unix socket when given a path instead of a port.
```
bartlomieju, line 15: Show example using `node:http` too (both server and client)

examples/scripts/http_proxy.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * @title Proxy HTTP requests
+ * @difficulty intermediate
+ * @tags cli, deploy
+ * @run -N <url>
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/API/Request} MDN: Request
+ * @resource {/examples/http_server} Example: HTTP Server: Hello world
+ * @group Network
+ *
+ * A reverse proxy receives requests and forwards them to another server.
```
bartlomieju, line 10: Give example of using `node:http` here too

examples/scripts/http_server_hot_reload.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @resource {/examples/http_server} Example: HTTP Server: Hello world
+ * @group Network
+ *
+ * Restarting a server by hand after every edit gets old fast. The
+ * --watch-hmr flag hot-replaces changed modules inside the running
+ * process, so the server keeps serving while you edit.
+ */
+
+// Nothing in the code needs to change; this is a plain HTTP server.
+Deno.serve(() => new Response("Hello, World!"));
```
bartlomieju: I think it would be better to split it into `Deno.serve(handler)` and then define `handler` before/after the server

examples/scripts/http_server_streaming.ts
```diff
@@ -13,7 +13,7 @@
 
 function handler(_req: Request): Response {
   // Set up a variable to store a timer ID, and the ReadableStream.
-  let timer: number | undefined = undefined;
+  let timer: ReturnType<typeof setInterval> | undefined = undefined;
```
bartlomieju, line 16: NodeJS.Timer?

examples/scripts/http_server_streaming.ts
```diff
@@ -40,5 +40,23 @@ function handler(_req: Request): Response {
...
+// turns any async iterable into a stream usable as a Response body.
+async function* ticks(): AsyncGenerator<Uint8Array> {
+  const encoder = new TextEncoder();
+  for (let i = 1; i <= 3; i++) {
+    yield encoder.encode(`tick ${i}\n`);
+    await new Promise((resolve) => setTimeout(resolve, 1000));
+  }
+}
+
+// deno-lint-ignore no-unused-vars
```
bartlomieju: We shouldn't need this lint directive here

examples/scripts/http_server_tls.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @title HTTP server: TLS
+ * @difficulty intermediate
+ * @tags cli
+ * @run -N -R <url>
+ * @resource {https://docs.deno.com/api/deno/~/Deno.ServeTlsOptions} Doc: Deno.ServeTlsOptions
+ * @resource {/examples/http_server} Example: HTTP Server: Hello world
+ * @group Network
+ *
+ * Deno.serve speaks HTTPS when given a certificate and private key. This
```
bartlomieju, line 10: Give example of using `node:https` too

bartlomieju:
Addressed the review comments:

- **fetch_unix_socket**: added the node:http pair, server via `listen(path)` and client via `request({ socketPath })` (verified: prints `Hello from node:http`).
- **http_proxy**: added a node:http variant of the proxy, bridging through fetch and streaming the body out via the Node response (verified against the upstream).
- **http_server_tls**: added a node:https section using the same certificate files (verified earlier with curl).
- **http_server_hot_reload**: split into a named `handler` + `Deno.serve(handler)`.
- **http_server_streaming**: the lint directive is gone; the generator handler is now actually served on `/ticks` via a tiny route, so both handlers are exercised (verified: `curl /ticks` streams `tick 1..3`).
- **NodeJS.Timer question**: on current Deno, `setInterval` returns the Node-compatible `Timeout` type even in files with no node imports, so the original `number` annotation no longer type-checks at all; `ReturnType<typeof setInterval>` is the version-portable spelling. Happy to switch to an explicit type if you prefer one.

Also swept the branch for em-dashes per the side request.

PR #3217, opened 2026-06-11
docs: add testing workflow examples and tutorials

bartlomieju:
Fills the task-shaped testing gaps in the examples catalog (follow-up to
the earlier example batches): skipping and focusing tests (the ignore
option and Deno.test.only, including the deliberate run failure that
guards against forgotten focus markers), filtering and controlling test
runs (--filter, --fail-fast, --watch, --parallel, and a sanitizer-safe
per-test timeout pattern since there is no built-in option), and using
Testing Library with happy-dom in deno test.

Existing pages were checked for overlap first: snapshot updating and fake
timers are already covered by the snapshot and mocking tutorials, and the
BDD tutorial covers describe/it, so those are linked rather than
duplicated. The Jest migration guide originally in this PR moved to a
separate branch. All test files run green and the Testing Library flow was
verified end to end before being written up.


runtime/test/migrate_from_jest.md
bartlomieju, line 1: I'm on the fence - wouldn't it be better to put it into the "Testing" section as a concrete page?

examples/tutorials/migrate_from_jest.md
```diff
@@ -0,0 +1,97 @@
...
+standard library provides the same `describe`/`it` structure, an `expect` with
+the matchers you already use, and mock functions. What changes is a handful of
+imports and how the runner is configured.
+
+## The same test, in Deno
+
+A typical Jest test needs only new imports:
+
+```ts title="cart.test.ts"
+import { beforeEach, describe, it } from "jsr:@std/testing/bdd";
```
bartlomieju: This should suggest using `node:test` instead of using `@std/testing/bdd`

examples/tutorials/testing_library.md
```diff
@@ -0,0 +1,73 @@
+---
+title: "Use Testing Library with Deno"
+description: "Test DOM behavior with @testing-library/dom and happy-dom in deno test: querying by role and text, simulating events, and the permissions the setup needs."
+url: /examples/testing_library_tutorial/
+---
+
+[Testing Library](https://testing-library.com/) encourages asserting on what the
+user sees — roles, labels, text — instead of implementation details. It runs in
```
bartlomieju:
```suggestion
user sees (roles, labels, text) instead of implementation details. It runs in
```

examples/tutorials/testing_library.md
```diff
@@ -0,0 +1,73 @@
...
+`deno test` with a simulated DOM from
+[happy-dom](https://github.com/capricorn86/happy-dom).
+
+## Install the packages
+
+```sh
+deno add npm:happy-dom npm:@testing-library/dom
```
bartlomieju:
```suggestion
deno add happy-dom @testing-library/dom
```

examples/tutorials/testing_library.md
```diff
@@ -0,0 +1,73 @@
...
+`getByRole("button")` fails the test if the element stops being a button, and
+`getByText` fails if the visible text changes — the assertions track what a user
+experiences rather than class names or DOM structure. The full query catalog
+(`findBy`, `queryAll`, label and placeholder queries) is in the
+[Testing Library documentation](https://testing-library.com/docs/queries/about).
+
+## Where this fits
+
+For testing DOM code without Testing Library — using `deno-dom` or parsing HTML
+directly — see [Testing web applications](/examples/web_testing_tutorial/).
```
bartlomieju:
```suggestion
For testing DOM code without Testing Library (using `deno-dom` or parsing HTML
directly) see [Testing web applications](/examples/web_testing_tutorial/).
```

bartlomieju:
Addressed the review:

- **Moved the Jest guide into the runtime Testing section** as a concrete page at `/runtime/test/migrate_from_jest/`, with a "Migrating from Jest" sidebar entry next to Testing (mirroring the "Migrating from Node" precedent). The examples catalog keeps a pointer card so it stays discoverable from /examples/ too.
- **`node:test` is now the primary mapping** for `describe`/`it`/hooks (verified: the snippet runs green under `deno test` with `@std/expect`), with `@std/testing/bdd` mentioned as the std alternative in the table. Also added a line that `node:test` suites keep working in Node itself, which is a nice migration-safety property.
- Applied all three suggestions on the Testing Library page (the first was already in after the em-dash sweep; `deno add` now uses unprefixed names; the parenthesized closing sentence is verbatim from the suggestion).

PR #3220, opened 2026-06-11
docs: split the Testing guide into a hub and six spoke pages

bartlomieju:
The testing guide was one 3,200-word page where the most-searched topics
were link farms: mocking, snapshots, and coverage had no code at all, and
the update workflow for snapshots was undocumented. The guide is now a hub
with six spokes under /runtime/test/: mocking and test doubles (absorbs the
mocking and stubbing tutorials, with the FakeTime example fixed to the
disposable pattern), snapshot testing (absorbs the snapshot tutorial and
makes the --update workflow prominent), test coverage (sample output pasted
from a real run; since deno coverage has no threshold flag, the page shows
honest CI gating patterns instead of inventing one), documentation tests
and sanitizers (moved out of the hub), and Migrating from Jest (lifted from
#3217 so that PR can stay focused on examples; its Testing Library link is
deferred until that PR's tutorial lands).

Absorbed tutorials redirect via oldUrl and the topic-specific legacy
redirects moved from the hub to the matching spoke (verified in the built
redirect map). The sidebar nests the spokes under Testing, and the examples
gallery and all inbound links point at the new pages.


bartlomieju:
Screenshot 2026-06-11 at 20 56 47

https://docs-dat2j6emgr0b.deno.deno.net/runtime/test/coverage/

I'm honestly not convinced this is a good idea - we already have it for "CLI reference" and "@std" and I find these nested lists confusing. Would it be better served as top-level section? I'd say yes, but the problem is that we lot of pages are gonna get split into multiple smaller packages (eg. #3223). Is there another way? Should we leverage topbar subnav like in https://docs-dat2j6emgr0b.deno.deno.net/deploy/ to have "tiered" sidebars?

josh-collinsworth:
> I'm honestly not convinced this is a good idea - we already have it for "CLI reference" and "@std" and I find these nested lists confusing. Would it be better served as top-level section? I'd say yes, but the problem is that we lot of pages are gonna get split into multiple smaller packages (eg. #3223). Is there another way? Should we leverage topbar subnav like in https://docs-dat2j6emgr0b.deno.deno.net/deploy/ to have "tiered" sidebars?

I don't think I would make Testing its own section. It seems to me like guides is where it belongs and if you start pulling individual features out, who knows where you stop.

I think I'd say if you expect to do this with _more_ subsections, go with this new approach. It's a little weird that Testing is the _only_ thing outside the Reference section that has a dropdown. If that's changing, and more things will be put into expandable containers, I'd say go with this new approach. Once it's a convention, it'll be less confusing.

Otherwise, I'd leave it the way it is now. Again: it's a little weird that it's almost the only thing with a dropdown, _and_ we make better use of the righthand space currently.

That's my two cents.

runtime/test/coverage.md
```diff
@@ -0,0 +1,194 @@
...
+  - /runtime/manual/basics/testing/coverage/
+---
+
+Code coverage tells you which lines, branches, and functions of your code
+actually ran while your tests executed. Deno has coverage built in: no
+instrumentation step, no extra dependencies. The workflow is two commands:
+`deno test --coverage` collects the data, and
+[`deno coverage`](/runtime/reference/cli/coverage/) turns it into a report.
+
+## Collect coverage data
```
bartlomieju, line 14: Ensure we mention collecting coverage in `deno run` (useful when spawning subprocesses, etc)

bartlomieju: Added in f5fa4a3e: a "Collect coverage outside deno test" subsection covering `deno run --coverage` and the subprocess case via `DENO_COVERAGE_DIR` (child processes inherit the env var, so their execution lands in the same report). Both verified by running them — including a test that spawns a Deno subprocess and confirms the child's file appears in the combined report.

PR #3221, opened 2026-06-11
docs: add WebSocket, HTML extraction, and runtime utility examples

bartlomieju:
Closes the remaining small task-shaped gaps in the examples catalog:
broadcasting WebSocket messages to all connected clients (with the
single-process caveat and pointers to shared backends), attaching
per-socket state to WebSocket connections, extracting links and Open
Graph metadata from fetched HTML with deno-dom, pinning the process time
zone with TZ, taking V8 heap snapshots programmatically via node:v8 for
DevTools analysis, and finding an executable on the PATH (a small which
implementation with the unix/Windows differences noted).

The existing Docker reference guide is now also surfaced as a card in the
examples catalog's deployment section, where people actually look for it.
One scoping note: WebSocket compression got a one-line honesty note in
the per-socket-state example instead of a page, since the upgrade does
not negotiate per-message deflate (verified: negotiated extensions come
back empty).

Both WebSocket servers were verified with live clients, and every
standalone script runs with outputs matching its comments.

### DIFF
--- a/examples/scripts/set_timezone.ts
+++ b/examples/scripts/set_timezone.ts
@@ -1,0 +1,38 @@
+/**
+ * @title Set the time zone
+ * @difficulty beginner
+ * @tags cli
+ * @run <url>
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat} MDN: Intl.DateTimeFormat
+ * @resource {/examples/temporal/} Example: Temporal API
+ * @group System
+ *
+ * Date, Intl, and Temporal all format times in the process time zone,
+ * which defaults to the operating system setting. The TZ environment
+ * variable overrides it, which makes time-dependent behavior reproducible.
+ */
+
+// The current time zone is visible through the Intl API.
+const zone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+console.log(zone); // Europe/Warsaw (or whatever the OS is set to)
+
+// Everything that formats local time follows it.
+console.log(new Date().toString());
+
+// Start the process with TZ set to pin the zone, for example to keep
+// timestamps in CI logs consistent or to reproduce a user's bug report:
+//
+//   TZ=Asia/Tokyo deno run main.ts
+//   Asia/Tokyo
+//   Thu Jun 11 2026 22:15:00 GMT+0900 (Japan Standard Time)
+//
+//   TZ=UTC deno run main.ts
+//   UTC
+
+// To format a single value in another zone without changing the process
+// zone, pass timeZone to Intl instead.
+const tokyo = new Intl.DateTimeFormat("en-US", {
+  timeStyle: "long",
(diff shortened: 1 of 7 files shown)


examples/scripts/set_timezone.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * @title Set the time zone
+ * @difficulty beginner
+ * @tags cli
+ * @run <url>
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat} MDN: Intl.DateTimeFormat
+ * @resource {/examples/temporal/} Example: Temporal API
+ * @group System
+ *
+ * Date, Intl, and Temporal all format times in the process time zone,
```
bartlomieju, line 10: add Temporal example too (it should be the preferred one now)

examples/scripts/websocket_context.ts
```diff
@@ -0,0 +1,60 @@
...
+// pull it from: query parameters, cookies, or an auth header.
+interface Session {
+  username: string;
+  joined: Date;
+}
+
+const sessions = new Map<WebSocket, Session>();
+
+Deno.serve((req) => {
+  const { socket, response } = Deno.upgradeWebSocket(req);
```
bartlomieju, line 28: Could use validation like websocket_pubsub.ts

examples/scripts/websocket_pubsub.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// Connect two clients from another terminal and every message appears in
+// both:
+//
+//   const ws = new WebSocket("ws://localhost:8000/");
+//   ws.onmessage = (event) => console.log(event.data);
+//   ws.onopen = () => ws.send("hello, everyone");
+
+// This state lives in one process. To broadcast across several servers or
+// deno serve --parallel workers, relay messages through something shared,
```
bartlomieju: BroadcastChannel could work here too I think (verify and update if needed)

bartlomieju:
Addressed all three comments, plus resolved the conflicts with main (entries translated to the new flat _data.ts format; the heap snapshot example moved from the dissolved Advanced category to CLI):

- **set_timezone**: Temporal is now presented as the preferred API: Temporal.Now.zonedDateTimeISO() carries the zone id (follows TZ), and per-zone values are an explicit argument rather than a process setting; the Intl variant stays as the formatting path. Runs without flags.
- **websocket_context**: gained the same upgrade-header validation as the pubsub example (verified: plain GET now returns 426, WebSocket flow unchanged).
- **websocket_pubsub / BroadcastChannel**: verified empirically before editing. Result: BroadcastChannel does NOT span separate local processes (a second deno process receives nothing), but DOES reach the other workers of deno serve --parallel (workers received messages they had not sent) and other instances on Deno Deploy. The closing comment now shows the BroadcastChannel relay pattern for --parallel/Deploy and keeps KV watch/Redis for separate processes and machines.

PR #3222, opened 2026-06-11
docs: add import maps tutorial and surface debugging guide

bartlomieju:
Two small catalog gaps: re-mapping import paths had no task page despite
being the tsconfig-paths question every migrating project asks (the new
tutorial covers directory aliases like @/, single-module bare names, editor
support, and scoped overrides, all verified in a sample project), and the
existing debugging guide was invisible from the examples catalog, so it now
has a pointer card in the Advanced section like the Docker guide.

Note: the import maps tutorial links to the add/remove dependencies
tutorial from the open package management PR; that link goes live when
both are merged.

### DIFF
--- a/examples/tutorials/import_maps.md
+++ b/examples/tutorials/import_maps.md
@@ -81,26 +81,14 @@ for imports that originate under a given path prefix:
 
 Modules under `vendor/` get the legacy logger; everything else gets the default.
 
-## Overriding versions in package.json projects
+:::note
 
-The import map redirects specifiers in your own code. To force a version of a
-package deeper in the dependency tree, projects with a `package.json` use its
-`overrides` field, which Deno honors the same way npm does:
+The import map redirects specifiers in your own code. To force the version of a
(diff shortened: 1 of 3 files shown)


examples/tutorials/import_maps.md
bartlomieju, line 1: we should add example of using package.json `overrides` too

bartlomieju:
Re the overrides suggestion: rather than teaching overrides twice, the import maps tutorial now stays focused on specifier remapping and carries a note distinguishing the two mechanisms, linking to the canonical coverage that already exists in [Overriding transitive dependencies](https://docs.deno.com/examples/add_remove_dependencies_tutorial/#overriding-transitive-dependencies) (landed in #3214). That section gained reverse links too: to the import maps tutorial for own-import remapping and to the runtime [Overriding dependencies](https://docs.deno.com/runtime/packages/#overriding-dependencies) reference for the links-field/local-copy flavor. So the three override mechanisms (import map remapping, package.json overrides, links) are each taught once and cross-linked.

PR #3224, opened 2026-06-11
docs: rewrite the web development guide

bartlomieju:
This rewrites the web development guide so it actually guides, instead of
listing scaffold commands.

The page is structured around the three ways to build for the web with
Deno, introduced by a decision table so a reader can route themselves: no
framework, a Deno-native framework, or an npm framework.

The "start without a framework" section leads with a real, verified server
(an HTML page, a JSON API route, and static files in about twenty lines)
built on `Deno.serve` with standard `Request`/`Response`. It shows minimal
permissions (`--allow-net --allow-read`, only what it touches) and then
points to the HTTP server guide for routing, streaming, WebSockets, and the
rest. Hono and Oak live here as the lightweight-router step up, which this
page owns (the HTTP server guide links here).

Fresh and Lume stay as full Deno-native sections, since they are genuinely
different (no `package.json`, config in `deno.json`/`_config.ts`).

The npm frameworks are no longer six near-identical sections. They collapse
into one explanation of the shared pattern (scaffold with `-A`, `deno
install` to create `node_modules`, run with `deno task`) plus a
verified-command table. Folding them into a table makes room to cover more
frameworks without bulk: alongside Next.js, Astro, Vite, and Docusaurus, the
table now includes SvelteKit and Nuxt, both freshly scaffolded under Deno to
confirm the command and the resulting project shape (Nuxt's generator even
lists Deno as a package manager). Per-framework caveats (Vite templates,
Astro's install wizard, Docusaurus `--skip-install`, the interactive
SvelteKit and Nuxt prompts) are kept as short notes.

The built-in-server snippet and all six framework setups were executed
under Deno before being documented. URL and redirect entries are unchanged.


runtime/fundamentals/web_dev.md
```diff
@@ -1,119 +1,124 @@
...
-   allowing you to write TypeScript code without additional configuration or
-   tooling.
-3. Deno comes with a [standard library](/runtime/reference/std/) that includes
-   modules for common tasks like HTTP servers, file system operations, and more.
+A note on permissions before you start: the scaffolding commands below use `-A`
+(allow all permissions) because project generators genuinely need broad access.
+They write files, read environment variables, and download and execute packages.
+Servers you write yourself need far less, and the
+[no-framework section](#without-a-framework) shows what minimal permissions look
+like in practice.
```
bartlomieju: Add a note that if user doesn't want to grant all permission they can run without a flag and answer interactive prompts

PR #3226, opened 2026-06-11
docs: add server pattern examples

bartlomieju:
First batch of the catalog expansion toward 300 examples: the everyday
server patterns that had no task pages despite steady search demand. CORS
with a real origin allowlist, cookies via @std/http, server-side sessions
with expiry, basic auth with timing-safe credential comparison, token
bucket rate limiting with retry-after, graceful shutdown draining
in-flight requests on SIGTERM, caching headers with ETag/304 handling,
per-request timeouts returning 504, health and readiness endpoints,
offset and cursor pagination, webhook HMAC signature verification, RSS
feed generation, sitemap.xml and robots.txt, downloading files with
progress, routing fetch through an HTTP proxy (including the
HTTP_PROXY/NO_PROXY env vars, verified against the docs and empirically),
and finding a free port.

Every server was verified with real requests and the outputs in the
comments are captured, not invented, including the webhook flow (a signed
and a tampered payload) and the proxy example (a tiny logging proxy
observing the forwarded request). Also adds the existing
creating_and_verifying_jwt example to the sidebar, which was published
but missing from the catalog.


examples/scripts/download_progress.ts
bartlomieju, line 1: This one might be already covered by another example, please check

PR #3229, opened 2026-06-11
docs: reorganize example categories and add category quick-nav

bartlomieju:
Three structural fixes to the examples catalog. The FileSystem category is
renamed to "File system". The grab-bag Advanced category is dissolved into
real homes: file system events to File system, module metadata to Modules,
file-based routing and the WebSocket apps to Network, the word finder to
Web frameworks, and the AI items into a new AI category. The Unstable APIs
category is dissolved too, since most of its contents are no longer
unstable: text and byte imports join their JSON sibling in Encoding, the
UDP pings join the TCP ones in Network, and KV, KV watch, queues, and cron
form a new "Deno KV and scheduling" category. A "Standard library"
category is seeded with the collections and backoff items and will grow
with the upcoming std-focused examples.

The landing page also gains a category quick-nav: a row of anchor pills
under the type filters linking to each section, so the 18 categories are
one click away instead of a scroll hunt. No pages move, only sidebar
grouping and labels, so there is no URL or SEO impact. Item count is
conserved (223 before and after on this branch).

Note for sequencing: this touches many section boundaries in
examples/_data.ts, so it will conflict trivially with the other open
example PRs; happy to rebase it last.

### DIFF
--- a/_components/LandingPage.tsx
+++ b/_components/LandingPage.tsx
@@ -179,6 +179,21 @@ export default function LandingPage(
           </span>
         </section>
 
+        <nav
+          className="flex flex-wrap gap-2 mb-10"
+          aria-label="Example categories"
+        >
+          {sidebar.map((category: { title: string }) => (
+            <a
+              key={category.title}
+              href={`#${category.title.toLowerCase().replace(/\s+/g, "-")}`}
+              className="px-3 py-1 rounded-full border border-foreground-tertiary text-sm !no-underline text-foreground-secondary hover:border-primary hover:text-primary transition-colors duration-150"
+            >
+              {category.title}
+            </a>
+          ))}
+        </nav>
+
         <section className="unfiltered mb-8 w-full markdown-body">
           {componentsPerSidebarItem}
         </section>
--- a/examples/_data.ts
+++ b/examples/_data.ts
@@ -129,6 +129,7 @@ export const sidebar = [
       },
     ],
   },
+
   {
     title: "Modules and package management",
     items: [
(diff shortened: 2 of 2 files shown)


bartlomieju:
Added strategic category ordering on top of the reorganization. The scheme is user journey first, demand breaking ties:

1. **Learn**: Basics, Modules and package management
2. **Core runtime tasks by demand**: Network, File system, System, Web standard APIs, Standard library, Encoding, CLI, Cryptography
3. **Quality**: Testing
4. **Build applications**: Web frameworks and libraries, Databases, Deno KV and scheduling, AI
5. **Ship and operate**: Deploying Deno projects, OpenTelemetry, Deno Sandbox

OpenTelemetry moves from seventh to third-from-last; Network and File system move into the top tier. The quick-nav pills and sidebar inherit the order automatically. A future refinement could be reordering based on page analytics, but this gets the obvious cases right today. (The stacked #3230 is already merged up with this change; item counts conserved at 223 here, 241 there.)

bartlomieju:
Refactored the data model so ordering is no longer encoded in array position: examples/_data.ts now declares a flat `items` list (each item carries its `category`), a small `categoryOrder` array that controls section order, and `sidebar` is derived from both with items alphabetized within each category at build time. Consumers are untouched since the derived export has the same shape.

Practical effects: reordering categories is now a one-line array edit instead of a 600-line diff, adding an example is a single append anywhere in the flat list, and concurrent example PRs stop conflicting structurally. One behavior change to be aware of: within-category order is now strictly alphabetical, so e.g. Basics no longer leads with "What is Deno?" — if some categories deserve a curated lead item, an optional pin field would be a tiny follow-up.

PR #3230, opened 2026-06-11
docs: add standard library and encoding examples

bartlomieju:
Second batch of the catalog expansion: task pages for the standard
library's everyday helpers and the remaining encoding formats. The new
Standard library category (from the category reorganization this PR is
stacked on) gains debounce, concurrency-limited task pools, promise
timeouts, memoization with LRU eviction, semver parsing and comparison,
string case conversion, regexp escaping, date formatting and parsing,
structured logging with file handlers, and human-readable bytes and
durations. Encoding gains INI, JSONC, streaming CSV, JSON Lines,
MessagePack, CBOR, and node:zlib compression, and File system gains tar
archive creation and extraction with the current TarStream API.

Every script runs with outputs verified against the comments, which
surfaced several facts worth having in the docs rather than guessed:
pooledMap yields results in input order rather than completion order,
deadline rejects with a TimeoutError DOMException, the binary byte format
prints kiB with a lowercase k, ini values parse as strings unless a
reviver converts them, CBOR round-trips Date values while MessagePack
rejects them, and the tar module's current API is TarStream/UntarStream
rather than the older Tar/Untar classes.

Stacked on #3229 because it registers items into the new Standard library
category; it should merge after it.


examples/scripts/csv_streaming.ts
bartlomieju, line 1: verify this doesn't duplicate already existing example

examples/scripts/date_format_parse.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @run <url>
+ * @resource {https://jsr.io/@std/datetime/doc/~/format} Doc: @std/datetime format
+ * @resource {https://jsr.io/@std/datetime/doc/~/parse} Doc: @std/datetime parse
+ * @resource {https://docs.deno.com/examples/temporal/} Example: Temporal API
+ * @group Standard library
+ *
+ * The @std/datetime package formats and parses plain Date objects with
+ * pattern strings like yyyy-MM-dd, and computes human friendly
+ * differences between two dates. For richer needs such as time zones
+ * and calendar arithmetic, the Temporal API is the modern option.
```
bartlomieju: Meh, sort of skeptical, couldn't Temporal be used for that? If so this example should be rewritten to use temporal instead of std

bartlomieju:
Addressed both comments, retargeted the PR base to main (its old base #3229 merged), and resolved conflicts:

- **csv_streaming duplication check**: not a duplicate of the existing CSV example, and the differentiation was already wired in: parsing_serializing_csv covers the in-memory parse/stringify calls, csv_streaming covers CsvParseStream for files that should not be loaded whole, and each links to the other (resource link + an intro sentence naming the sync example).
- **date_format_parse**: rewritten Temporal-first as suggested, and moved from the Standard library category to Web standard APIs accordingly. It now covers PlainDate/PlainDateTime parsing with overflow: reject validation, toLocaleString formatting, calendar-aware arithmetic with add/until/total, and calendar properties; @std/datetime survives only in a final section for parsing non-ISO custom patterns, where Temporal has no equivalent. Verifying the rewrite caught a real timezone bug in that last section: std's parse interprets the input in the local zone, so converting to Temporal via UTC shifted the calendar date back a day; the example now routes through Temporal.Now.timeZoneId() and documents why.

PR #3231, opened 2026-06-11
docs: add web platform API and cryptography examples

bartlomieju:
Third batch of the catalog expansion: web platform APIs that Deno supports
but had no example, and a rounded-out cryptography section. Web standard
APIs gains URLPattern routing, AbortController and fetch timeouts (with
the real TimeoutError/AbortError/TypeError distinction), structuredClone
(including what does not clone), FormData round trips, EventTarget custom
events, BroadcastChannel between workers, TransformStream pipelines, the
full Intl family (numbers and currencies, dates, relative time, plurals
and lists, segmentation with grapheme-safe truncation), performance marks
and measures, and the Web Cache API (verified working in Deno with no
permission flags, proven against a request-counting local server).
Cryptography gains ECDSA signing, deriving AES keys from passwords with
PBKDF2, secure random values (with the modulo-bias explanation),
constant-time comparison, hashing large files as streams, and node:crypto
for npm compatibility.

All 21 scripts run with outputs verified against their comments; the
streaming hash example is additionally corroborated by a NIST SHA-256
test vector, and the Intl outputs were captured byte-for-byte from the
runtime's ICU rather than copied from browser documentation.


examples/scripts/intl_date_format.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat} MDN: Intl.DateTimeFormat
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange} MDN: formatRange
+ * @group Web Standard APIs
+ *
+ * Intl.DateTimeFormat turns a Date into text for any locale and time
+ * zone. The dateStyle and timeStyle shortcuts cover most needs, and
+ * formatRange renders a span of two dates without repeating the shared
+ * parts. This example formats one fixed instant in several ways.
+ */
+
```
bartlomieju, line 15: Check if this is not covered already, might also be worth adding Temporal examples here

examples/scripts/intl_relative_time.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @run <url>
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat} MDN: Intl.RelativeTimeFormat
+ * @resource {https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/format} MDN: format
+ * @group Web Standard APIs
+ *
+ * Intl.RelativeTimeFormat produces phrases like 3 days ago or in 2
+ * months for any locale. You pass a signed number and a unit, and the
+ * formatter handles grammar, pluralization, and translation. This
+ * example also builds a small helper that picks the right unit from a
+ * millisecond difference.
```
bartlomieju, line 14: Check if this is covered, also check if Temporal could/should be added here

examples/scripts/timing_safe_comparison.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @group Cryptography
+ *
+ * An ordinary comparison like a === b returns as soon as the first byte
+ * differs, so a correct first byte takes measurably longer to reject than
+ * a wrong one. By timing many requests, an attacker can recover a secret
+ * one byte at a time. When comparing API keys, tokens, or digests, use a
+ * comparison whose duration does not depend on where the inputs differ.
+ */
+
+import { timingSafeEqual } from "jsr:@std/crypto/timing-safe-equal";
```
bartlomieju, line 17: Also show examples from @std/crypto

bartlomieju:
Addressed all three comments (and resolved conflicts with main):

- **intl_date_format**: checked for overlap first: the @std/datetime example (pattern-string formatting, in the open std-library PR) and the existing Temporal API example cover different ground, so nothing is duplicated. Added a Temporal section showing toLocaleString takes the same locales and options on an Instant, and that a ZonedDateTime carries its zone (the timeZone option is neither needed nor allowed there, verified: passing one throws). Outputs captured from real runs.
- **intl_relative_time**: same overlap check, no duplication. Added a Temporal section where since() on zoned values produces the signed value for the formatter from calendar math instead of fixed millisecond constants (verified: 3 days ago). One API note from verification: Instant.since only supports time units, so the example goes through toZonedDateTimeISO for day-level math.
- **timing_safe_comparison**: the example was already built on timingSafeEqual from @std/crypto; it now also showcases the module's extended WebCrypto, using a BLAKE3 digest (verified, 32 bytes) feeding the same comparison pattern.

PR #3232, opened 2026-06-11
docs: add system, file system, and CLI tooling examples

bartlomieju:
Fourth batch of the catalog expansion: process and OS tasks, file system
operations, and the CLI tooling that only existed as videos. System gains
piping into subprocesses, custom env/cwd for children, TTY detection and
terminal size, OS information, network interfaces, exit codes, Node-style
worker_threads (with the parentPort.close gap documented and worked
around), EventEmitter, and node:readline. File system gains metadata
(stat/chmod/utime), glob matching, cross-process file locking (proven
with a child process blocking until the parent releases), truncation, and
hard links. CLI gains deno compile (including a real cross-compilation to
Linux from macOS), Deno.bench with a real results table, deno outdated
with --update vs --latest semantics, progress bars and spinners from
@std/cli with TTY-aware fallbacks, raw ANSI terminal control, and a
subcommand dispatch pattern with parseArgs.

All 20 scripts run with their declared permission flags and outputs
captured from real runs, including pty-based verification for the
terminal-dependent examples. Permission flags were determined empirically
(for example, networkInterfaces needs --allow-sys, not --allow-net).


examples/scripts/cli_subcommands.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @resource {/examples/command_line_arguments} Example: Command line arguments
+ * @group CLI
+ *
+ * Tools like git and deno itself take a subcommand first and flags after
+ * it. The pattern is easy to build by hand: treat the first element of
+ * Deno.args as the command name, dispatch to a function, and let each
+ * command parse its own flags with parseArgs from the standard library.
+ */
+
+import { parseArgs } from "jsr:@std/cli@^1.0.30/parse-args";
```
bartlomieju: This example should start by showing `parseFlags` from `node:util` first

examples/scripts/file_locking.ts
```diff
@@ -0,0 +1,69 @@
...
+await file.unlock();
+file.close();
+`;
+await Deno.writeTextFile(`${dir}/child.ts`, childScript);
+
+// The main process takes the exclusive lock first. Passing true means
+// exclusive (one holder); without it the lock is shared, which lets many
+// readers coexist while still excluding writers.
+const file = await Deno.open(path, { read: true, write: true });
+await file.lock(true);
```
bartlomieju, line 40: also showcase `tryLock` method

examples/scripts/glob_patterns.ts
```diff
@@ -0,0 +1,61 @@
...
+await Deno.writeTextFile(`${dir}/src/main.ts`, "");
+await Deno.writeTextFile(`${dir}/src/util/helpers.ts`, "");
+await Deno.writeTextFile(`${dir}/src/util/helpers_test.ts`, "");
+await Deno.writeTextFile(`${dir}/build/bundle.ts`, "");
+await Deno.writeTextFile(`${dir}/readme.md`, "");
+
+// expandGlob walks the file system and yields every entry that matches the
+// pattern. The ** segment matches any number of directories. The exclude
+// option filters out paths, here the generated build directory.
+import { expandGlob } from "jsr:@std/fs";
```
bartlomieju, line 29: Let's showcase `glob` from `node:fs`

examples/scripts/network_interfaces.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Deno.networkInterfaces lists every network address of the machine. The
+ * most common use is finding the LAN address of the computer, for example
+ * to print a URL that teammates on the same network can open. The call
+ * needs the system permission, granted with -S (short for --allow-sys).
+ */
+
+// Each entry describes one address bound to one interface. An interface
+// like en0 typically appears several times, once per IPv4 and IPv6 address.
+const interfaces = Deno.networkInterfaces();
```
bartlomieju, line 18: Also show `node:net` API

examples/scripts/subprocess_env_cwd.ts
```diff
@@ -0,0 +1,46 @@
...
+// With clearEnv set to true the child does not inherit anything, it only
+// gets the variables listed in env. Even PATH is gone, so this is best
+// combined with an absolute path to the executable. Note that the operating
+// system may still inject a few variables of its own.
+const cleared = await new Deno.Command(Deno.execPath(), {
+  args: ["eval", "console.log('has PATH:', Deno.env.has('PATH'))"],
+  env: { GREETING: "hello again" },
+  clearEnv: true,
+}).output();
+console.log(decoder.decode(cleared.stdout).trim()); // has PATH: false
```
bartlomieju: Use helpers we added in https://deno.com/blog/v2.5#simpler-stdio-from-denochildprocess also shorthand method for spawning too https://deno.com/blog/v2.7#new-apis-for-subprocesses

examples/scripts/subprocess_stdin.ts
```diff
@@ -0,0 +1,55 @@
...
+await writer.write(encoder.encode("cherry\n"));
+
+// Closing the writer closes the child's stdin. This is important: sort waits
+// for the end of its input before it prints anything, so forgetting this
+// step would hang the program.
+await writer.close();
+
+// output() waits for the process to exit and collects its stdout.
+const { stdout } = await sortProcess.output();
+const sorted = new TextDecoder().decode(stdout);
```
bartlomieju: Use helper method to get stdout (from previous commend)

examples/scripts/subprocess_stdin.ts
```diff
@@ -0,0 +1,55 @@
...
+console.log(sorted.trim().split("\n")); // [ "apple", "banana", "cherry" ]
+
+// When the input already exists as a stream or an iterable, skip the manual
+// writer and pipe it into stdin in one shot. ReadableStream.from turns any
+// iterable into a stream, and pipeTo closes the child's stdin when the
+// stream ends.
+const catProcess = new Deno.Command("cat", {
+  stdin: "piped",
+  stdout: "piped",
+}).spawn();
```
bartlomieju: Use `Deno.spawn()`

examples/scripts/terminal_detection.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @run <url>
+ * @resource {https://docs.deno.com/api/deno/~/Deno.stdout} Doc: Deno.stdout
+ * @resource {https://docs.deno.com/api/deno/~/Deno.consoleSize} Doc: Deno.consoleSize
+ * @group System
+ *
+ * A program can check whether it is talking to a real terminal or to a pipe.
+ * This is how command line tools decide to print colors and progress bars
+ * for humans, but plain machine readable text when their output is piped to
+ * another program or a file.
+ */
```
bartlomieju, line 14: also showcase using `process.stdout.isTTY`

bartlomieju:
Addressed all eight comments, with the new subprocess APIs verified on the current runtime before use:

- **subprocess_stdin / subprocess_env_cwd**: now use the Deno 2.5 stdio helpers (child.stdout.text() instead of decode-by-hand) and the 2.7 spawn shorthand: Deno.spawn(cmd, args, opts) for the streaming cases and Deno.spawnAndWait for run-to-completion. One behavior worth noting in review: Deno.spawn defaults stdio to inherit, so the examples pass stdout: "piped" explicitly where the helpers are used. Deno.Command stays mentioned as the configure-then-spawn API.
- **file_locking**: tryLock showcased after the blocking story: the parent re-acquires with tryLock(true) and the comment frames it as the skip-if-busy pattern.
- **glob_patterns**: node:fs/promises glob added with cwd and exclude options, same matches as expandGlob (verified output).
- **network_interfaces**: the Node equivalent added. Small correction to the comment: the listing API lives in node:os (os.networkInterfaces(), a record keyed by interface name) rather than node:net, which only does sockets, so the example shows node:os.
- **terminal_detection**: process.stdout.isTTY added with its piped-vs-terminal values (undefined when piped, matching Node).
- **cli_subcommands**: converted to lead with parseArgs from node:util throughout (options + defaults per command), with a closing note pointing at @std/cli parseArgs for aliases/collect/negatable when a tool outgrows it.

All seven scripts re-run with their declared flags; outputs match.

bartlomieju:
Addressed all eight comments (plus resolved conflicts with main):

- **subprocess examples**: now use the new shorthand and stdio helpers, verified against the current runtime: Deno.spawn(cmd, args, opts) for spawning, child.stdout.text() for collection (Response-style helpers from 2.5), and Deno.spawnAndWait for run-to-completion. One behavior note discovered while verifying: Deno.spawn defaults stdout to inherit, so the examples pass stdout: "piped" explicitly before using the helpers.
- **cli_subcommands**: leads with parseArgs from node:util throughout (positionals + per-command options), with @std/cli parseArgs mentioned at the end for when a tool needs aliases/collect/negatable flags.
- **file_locking**: added tryLock with the skip-if-busy framing (returns false instead of blocking; verified true after the child releases).
- **glob_patterns**: added glob from node:fs/promises with cwd and exclude, output matching the expandGlob section.
- **network_interfaces**: added the Node API equivalent; note it lives in node:os (os.networkInterfaces(), keyed by interface name) since node:net has no interface listing.
- **terminal_detection**: added process.stdout.isTTY (undefined when piped, true in a terminal).

Also caught during the conflict resolution: the new benchmark example collided with the existing deno_bench video page at /examples/deno_bench/, so the script is renamed to /examples/write_benchmarks/ and both pages now build. All seven touched scripts re-run with verified output, types check without casts (spawn/tryLock are typed on the current runtime).

PR #3235, opened 2026-06-11
docs: add a Migrating from Bun guide

bartlomieju:
The migrate section's direction is "migrating TO Deno from anywhere", but it
only covered Node. This adds /runtime/migrate/migrate_from_bun/ for the Bun
audience: run an existing project (leading with the permission-prompt
difference, the one thing Bun users won't expect), a verified command
cheatsheet, API equivalents for the big surfaces (Bun.serve to Deno.serve,
bun:sqlite to node:sqlite, Bun.file, bun:test, Bun.$), and an honest section
on what has no direct equivalent (macros, HTMLRewriter).

Every disputed fact was checked rather than assumed: node:sqlite confirmed
by importing it, HTMLRewriter confirmed absent the same way, deno bundle
footnoted as experimental per its own page, Bun-side claims verified against
bun.com/docs, and bunx maps to deno x (per our own CLI reference) rather
than a deno run incantation. The page is deliberately not yet linked from
the migrate hub or sidebar; that wiring lands with tomorrow's sidebar
regroup to avoid conflicting with the PRs already in flight.

### DIFF
--- a/runtime/migrate/migrate_from_bun.md
+++ b/runtime/migrate/migrate_from_bun.md
@@ -1,0 +1,220 @@
+---
+title: "Migrating from Bun"
+description: "How to move a Bun project to Deno: run your package.json scripts unchanged, map Bun CLI commands to Deno, and replace Bun-specific APIs with Deno and Node built-ins."
+---
+
+Most Bun projects are standard `package.json` and TypeScript projects, and Deno
+runs those directly. Deno reads your `package.json`, installs the same npm
+dependencies, runs TypeScript without a build step, and ships a comparable
+built-in toolchain: a test runner, formatter, linter, bundler, and compiler in
+one binary.
+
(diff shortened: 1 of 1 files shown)


bartlomieju:
Closing in favor of a single combined PR that targets the `2.9` branch. Deno 2.9 adds automatic lockfile seeding (`deno install` seeds `deno.lock` from `package-lock.json`/`yarn.lock`/`pnpm-lock.yaml`, see denoland/deno#35330, #35350, #35346) and auto-migrates `pnpm-workspace.yaml` on resolution failure (#34993), so the "Deno does not read your lockfile / convert pnpm-workspace.yaml by hand" framing here is no longer accurate. The corrected npm/yarn/pnpm pages are carried into the new PR.

bartlomieju:
Superseded by #3353 (targets the `2.9` branch).

PR #3237, opened 2026-06-12
docs: opt-in sidebar disclosure for groups with landing pages

bartlomieju:
On #3220's preview, the "Testing" sidebar item has a landing page but is not
clickable: items with children render as an accordion button that swallows
the href. This adds a new, opt-in pattern instead of changing the existing
accordions: a group item with disclosure: true in its section's _data.ts
renders its label as a normal link (hover, active state, navigates to the
hub) with a separate chevron button that toggles the children. Opted-in
groups are independent of one another, the active trail is rendered open by
the server (no expand flash), everything else defaults to collapsed, and
manual toggles are remembered per group.

Nothing in this PR opts in, so the CLI reference and Standard library render
byte-for-byte as before (verified in the built output: legacy accordion
markup present, zero traces of the new pattern). #3220 adds disclosure: true
to its Testing item in a follow-up commit, which is where the new behavior
becomes visible. Visual polish is deliberately minimal, pending the styling
overhaul.

### DIFF
--- a/_components/SidebarNav/comp.tsx
+++ b/_components/SidebarNav/comp.tsx
@@ -26,7 +26,9 @@ export default function (data: Lume.Data) {
             <SidebarList>
               {nav.items?.map((item: any) => (
                 <li key={item.href}>
-                  {item.items && item.items.length > 0
+                  {item.items && item.items.length > 0 && item.disclosure
+                    ? <SidebarDisclosure item={item} currentUrl={currentUrl} />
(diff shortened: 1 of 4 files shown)


bartlomieju:
Folded into #3220 at the author's request so the Testing UX ships with the split itself (same commits, cherry-picked, plus the disclosure flag on the Testing item). The mechanism stays opt-in: CLI reference and Standard library render exactly as before.

PR #3240, opened 2026-06-12
docs: cover stdin, exit codes, prompts, and assets in the CLI apps guide

bartlomieju:
The CLI apps guide stopped at argument parsing and compilation, skipping the
things every real CLI needs next. It now covers reading piped stdin and
branching on Deno.stdin.isTerminal(), exit codes (including Deno.exitCode
for set-failure-but-keep-running, verified by running it: execution
continues and the process exits 1), interactive prompts (the built-in
prompt/confirm globals plus @std/cli's Spinner, labeled unstable because it
ships from the unstable-spinner module), embedding assets in compiled
binaries with --include plus import.meta.dirname (compiled and executed
from a different working directory to prove the lookup works), and a short
distribution note pointing at the compile reference's targets and
code-signing sections. Every snippet was executed before being documented.

### DIFF
--- a/runtime/cli_apps/index.md
+++ b/runtime/cli_apps/index.md
@@ -1,6 +1,6 @@
 ---
 title: "Build CLI apps"
-description: "Build command-line tools with Deno: read arguments, compile to a single self-contained executable, cross-compile for other platforms, and distribute your tool."
+description: "Build command-line tools with Deno: read arguments and stdin, prompt the user, set exit codes, compile to a single self-contained executable, and distribute your tool."
 ---
 
 Deno is a great way to ship command-line tools. Your tool is just TypeScript, so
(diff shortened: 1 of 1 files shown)


runtime/cli_apps/index.md
```diff
@@ -30,6 +30,122 @@ $ deno run greet.ts --name=Deno
...
+
+When your tool can be used both ways, use
+[`Deno.stdin.isTerminal()`](/api/deno/~/Deno.stdin.isTerminal) to tell whether
+stdin is an interactive terminal or a pipe, and branch accordingly. This is also
+how tools decide whether to print colors and progress bars (interactive) or
+plain machine-readable output (piped). Here the tool falls back to asking the
+user when nothing is piped in:
+
+```ts title="count.ts"
+let input: string;
```
bartlomieju, line 51: Also mention `process.stdin`

runtime/cli_apps/index.md
```diff
@@ -30,6 +30,122 @@ $ deno run greet.ts --name=Deno
...
+Enter some text: hi
+Got 2 characters
+```
+
+Reading stdin requires no permission flags.
+
+## Prompt the user
+
+For simple interactive input, Deno ships the browser globals `prompt()`,
+`confirm()`, and `alert()`. They block until the user responds, and they need no
```
bartlomieju: expain in a parens what "they block" means

runtime/cli_apps/index.md
bartlomieju: Added in f27cbe74: a note that Node-style process.stdin (node:process or the global) works in Deno too, so ported CLIs and npm input libraries run unchanged. Verified by piping into a process.stdin script under deno run.

runtime/cli_apps/index.md
bartlomieju: Clarified in f27cbe74: they block until the user responds (execution pauses at that line and nothing else runs until input arrives).

PR #3242, opened 2026-06-12
docs: add a guide for switching your package manager to Deno

bartlomieju:
The migrate section's smallest adoption step had no page: using Deno purely
as the package manager for a project you keep running with Node. This adds
/runtime/migrate/switch_package_manager/ with a four-column mapping table
(npm, yarn, pnpm, Deno) covering install/add/remove/update/outdated/run/
audit/why/dlx/ci. Every Deno command was verified against --help and the
yarn/pnpm names against their official docs (modern Yarn's missing outdated
command is noted rather than papered over).

The behavioral claims come from experiments, not assumptions: deno install
against an npm-generated project leaves package-lock.json byte-identical,
creates deno.lock and an isolated-layout node_modules, and does not honor
package-lock version pins within a range (tested with a pinned transitive
version); deno add -D writes devDependencies in package.json projects. The
page also covers what's different (lifecycle scripts off by default, audit
--fix, minimum dependency age) and what to watch (node_modules layout
modes). Not yet linked from the migrate hub; that wiring lands with the
sidebar regroup alongside the Bun guide's.


bartlomieju:
Wired the page into the docs and removed the overlap with the migrate index (36778424):

- Added it to the runtime sidebar under **Migrating from Node** (label "Package manager") so it is reachable from the menu.
- Trimmed the migrate index's "Use Deno as your package manager" section to a short teaser that links here, instead of repeating the install commands. The page keeps the full npm/yarn/pnpm mapping, the supply-chain/lockfile detail, and the caveats, which the index never covered.
- Added the missing `last_modified` frontmatter field.

deno fmt clean and the sidebar label-length test passes.

PR #3244, opened 2026-06-12
docs: Node compatibility rewrite, version A (walkthrough)

bartlomieju:
One of three competing ground-up rewrites of the Node compatibility page,
opened as drafts for side-by-side review; only one will merge. This version
threads a single "health report" script through the entire page: it picks up
an npm package, a node: built-in, and the process global, moves into
package.json and deno task, hits a legacy CommonJS helper, and ends with a
Node-API addon (duckdb). Every failure a reader would hit is triggered for
real and shown with Deno's actual error text and the fix: the bare-import
hint, the missing-deno-install error, require-is-not-defined, and the
ignored-build-scripts to --allow-scripts flow. All snippets were executed on
Deno 2.8 before being documented; the private-registries section, linked
anchors, and oldUrl redirects are unchanged from the fact-checked baseline
in #3241. Compare with version B (lookup) and version C (task recipes).


bartlomieju:
Siblings for comparison: version B is #3245, version C is #3246.

bartlomieju:
Superseded by #3245, which combines the strongest parts of all three drafts: this version's real executed error transcripts (missing deno install, require-is-not-defined, the ignored-build-scripts to --allow-scripts flow) are folded into the inline troubleshooting notes there. The single-growing-script structure was dropped in favor of a task-oriented spine, which reads better for a reference page people land on mid-task.

PR #3246, opened 2026-06-12
docs: Node compatibility rewrite, version C (task recipes)

bartlomieju:
One of three competing ground-up rewrites of the Node compatibility page,
opened as drafts for side-by-side review; only one will merge. This version
organizes the page by what the reader is trying to do: each section is a
goal (use an npm package, use a Node built-in, run an existing project, make
CommonJS work, control node_modules, use native addons, get type
definitions), opening with a working example, then the rules, then an
inline "if it doesn't work" note so troubleshooting lives next to the task
instead of in a separate section. An honest "How compatible is it?" section
near the top carries the test-suite stat. New snippets were executed on
Deno 2.8; the private-registries section, linked anchors, and oldUrl
redirects are unchanged from the fact-checked baseline in #3241. Compare
with version A (#3244) and version B (#3245).

### DIFF
--- a/runtime/fundamentals/node.md
+++ b/runtime/fundamentals/node.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-05-20
+last_modified: 2026-06-12
 title: "Node and npm Compatibility"
 description: "Guide to using Node.js modules and npm packages in Deno. Learn about compatibility features, importing npm packages, and differences between Node.js and Deno environments."
 oldUrl:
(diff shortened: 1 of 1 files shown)


bartlomieju:
Superseded by #3245. This version's task-oriented section spine and inline 'if it doesn't work' notes are the backbone of the combined page; #3245 adds version B's at-a-glance compatibility table and HTTP-server hook on top, plus version A's real error transcripts in the troubleshooting notes.

PR #3247, opened 2026-06-12
docs: render example snippets as stacked prose and code blocks

bartlomieju:
The two-column example layout renders commentary as italic, right-aligned
margin notes beside the code. Feedback on the recent example batches was
that this reads strangely: the eye jumps between columns, the right
alignment makes paragraphs ragged on the left edge, and long explanations
get squeezed into a narrow strip.

This PR is one of two alternative reworks, alongside #3248 (pick one). It
switches the example pages to a single-column flow: each piece of
commentary renders as a regular paragraph directly above the code it
describes, in document order. Code now splits into visual blocks at prose
boundaries rather than at grid rows, so a run of uninterrupted code stays
one connected block. Multi-file examples keep their filename chips on the
first block of each file. The column-alignment padding hacks in the
position CSS are no longer needed and were removed.

Before (current two-column layout):

before

After (stacked):

stacked

Multi-file example (web workers):

multi-file

### DIFF
--- a/examples/_components/ExamplePage.tsx
+++ b/examples/_components/ExamplePage.tsx
@@ -46,17 +46,25 @@ export default function ExamplePage({ example }: Props) {
             class="flex flex-col gap-4 md:gap-0 example-content"
             key={file.name}
           >
-            {file.snippets.map((snippet, i) => (
-              <SnippetComponent
-                key={i}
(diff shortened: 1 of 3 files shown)


bartlomieju:
Sticking with current approach for now

PR #3248, opened 2026-06-12
docs: make example margin notes read as plain prose

bartlomieju:
The two-column example layout renders commentary as italic, right-aligned
margin notes beside the code. Feedback on the recent example batches was
that this reads strangely: the right alignment makes paragraphs ragged on
the left edge and the italics tire the eye over a long page.

This PR is the conservative counterpart to #3247 (pick one). It keeps the
two-column structure, so the at-a-glance density of the current pages is
unchanged, and only restyles the notes: plain upright text, left aligned,
secondary foreground color, relaxed line height, and a small top offset
to align the first line with the code beside it. The diff is a single
class change in SnippetComponent.

Before:

before

After (refined two-column):

refined


bartlomieju:
Sticking with current approach for now

PR #3249, opened 2026-06-12
docs: retire database video pages and harden quick examples

bartlomieju:
Three small quality passes on the examples catalog that were parked
during the recent expansion, all from the consolidation section of the
content backlog.

The Mongoose and Prisma video pages duplicated their tutorials: same
topic, two catalog entries, and the relabeled titles ("Prisma" next to
"Prisma (video)") made the redundancy obvious. The recordings also
predate the current tutorials and risk contradicting them, while the
written tutorials are the higher-quality source. So the standalone video
pages are removed and their old URLs redirect to the tutorials; the
videos remain reachable on the YouTube channel. The catalog drops from
316 to 314 items. Redis keeps both of its pages since the quick start
(connect and send commands) and the tutorial (build a caching layer)
cover different ground; they now cross-link each other instead.

The Postgres quick example previously hardcoded credentials and let
connection failures escape as a raw stack trace. It now reads the
standard PG* environment variables with sensible localhost fallbacks and
handles query failure by reporting the error code; the failure path was
verified against a closed port (prints "Query failed: ECONNREFUSED") and
the pool is closed in a finally block either way.

The project initialization tutorial showed bare commands with no
expected output and had drifted from what deno init actually generates
(it described the old "add function" template; the current template is a
small HTTP server with two tests, and the dev task includes
--allow-net). The tutorial now shows the real captured output of deno
init, deno task dev, and deno test, so readers can verify each step as
they follow along.

### DIFF
--- a/examples/scripts/postgres.ts
+++ b/examples/scripts/postgres.ts
@@ -17,7 +17,9 @@ import postgres from "npm:postgres";
 
 // Read the connection settings from the standard libpq environment
 // variables. Locally you might set PGDATABASE=test and nothing else;
-// in production a managed database provides all of them.
+// in production a managed database provides all of them. A common pattern
+// is to keep these in a .env file and load it with the --env-file flag,
+// for example: deno run --env-file -N -E main.ts
 const sql = postgres({
   host: Deno.env.get("PGHOST") ?? "localhost",
   port: Number(Deno.env.get("PGPORT") ?? 5432),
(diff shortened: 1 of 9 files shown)


examples/scripts/postgres.ts
```diff
@@ -3,29 +3,43 @@
  * @difficulty intermediate
  * @tags cli, deploy
  * @run -N -E <url>
+ * @resource {https://www.npmjs.com/package/postgres} postgres on npm
  * @group Databases
  *
  * Using the npm Postgres client, you can connect to a Postgres database
- * running anywhere.
+ * running anywhere. Connection settings belong in environment variables
```
bartlomieju, line 10: I think it might be worth mentioning `.env` file and `--env-file` flag here since it's a popular pattern when working with DBs

PR #3250, opened 2026-06-12
docs: rewrite the TypeScript fundamentals page

bartlomieju:
The TypeScript page never showed running a .ts file, buried the
execution-vs-checking mental model in a callout, and said nothing to its
most likely reader: a Node.js developer arriving with a tsc/tsx/ts-node
toolchain. It also taught a pattern that crashes on Deno 2 — the global
augmentation examples extend `interface Window`, which type-checks but
throws `ReferenceError: window is not defined` at runtime (verified),
since `window` was removed in Deno 2.

This rewrite leads with `deno run main.ts`, explains the strip-and-run
vs opt-in-checking split up front (including that Deno never emits,
pointing at `deno transpile`/`deno pack`), and adds a "Coming from
Node.js" section: a toolchain mapping table plus the three things Node
developers trip on — real file extensions in imports, full-language
support versus erasable-only type stripping, and emit options having
nothing to configure. The global augmentation examples now declare
globals with `var` and access them via `globalThis`; the JavaScript
section's "TSDoc" label is corrected to JSDoc with canonical syntax;
`deno check`'s directory and `--check-js` forms are documented; and the
browser/worker lib material is condensed into a pointer at the
configuration reference instead of duplicating it. Every command and
claim was verified against Deno 2.8. Heading anchors linked from other
pages and all `oldUrl` redirect entries are unchanged.

### DIFF
--- a/runtime/fundamentals/typescript.md
+++ b/runtime/fundamentals/typescript.md
@@ -13,8 +13,9 @@ oldUrl:
 ---
 
 TypeScript is a first-class language in Deno. Put TypeScript in a file and run
-it — there is no compiler to install, no `tsconfig.json` to write, and no build
-step:
+it. There is no compiler to install and no build step, and configuration is
+optional: a `tsconfig.json` works if you already have one, but you don't need to
+write one to get started.
 
 ```ts title="main.ts"
 function greet(name: string): string {
@@ -29,7 +30,9 @@ deno run main.ts
 ```
 
 Deno also ships the type checker: `deno check` plays the role of `tsc --noEmit`,
-with strict mode on by default.
+with strict mode on by default. The same checker runs inside your editor through
+the [Deno language server](/runtime/getting_started/setup_your_environment/), so
+type errors surface as you type.
 
 ## How TypeScript runs in Deno
 
@@ -42,16 +45,19 @@ Deno treats executing TypeScript and type-checking it as two separate concerns:
   TypeScript compiler over your code without executing it.
 
 This means **type errors do not stop your code from running unless you ask**.
-Checking a whole module graph is expensive, so Deno keeps it out of the hot
+Type-checking a whole module graph is expensive, so Deno keeps it out of the hot
 edit-run loop and lets you invoke it where it earns its cost: in your editor
 (via the Deno language server), in CI, and in `deno test`, which type-checks by
 default.
 
-A consequence worth knowing: Deno never emits JavaScript files. There is no
-`outDir`, no `dist/` directory, and no source map configuration — error stack
-traces point at your `.ts` sources directly. If you genuinely need `.js` output
-(for example, to ship a library to consumers running Node or a bundler), that is
-a separate tool: [`deno transpile`](/runtime/reference/cli/transpile/) or
+A consequence worth knowing: Deno does not write JavaScript next to your
+sources. The stripped output goes into an internal cache instead (you can see
+its location listed as "Emitted modules cache" in the output of `deno info`), so
+there is no `outDir`, no `dist/` directory, and no source map configuration to
+manage. Error stack traces still point at your `.ts` sources directly. If you
+genuinely need `.js` files on disk (for example, to ship a library to consumers
+running Node or a bundler), reach for a separate tool:
+[`deno transpile`](/runtime/reference/cli/transpile/) or
 [`deno pack`](/runtime/reference/cli/pack/).
 
 ## Coming from Node.js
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
   - /runtime/fundamentals/
 ---
 
-TypeScript is a first class language in Deno. You can run or import TypeScript
-without installing anything more than the Deno CLI. With its built-in TypeScript
-compiler, Deno will compile your TypeScript code to JavaScript with no extra
-config needed. Deno can also type check your TypeScript code, without requiring
-a separate type checking tool like `tsc`.
+TypeScript is a first-class language in Deno. Put TypeScript in a file and run
+it — there is no compiler to install, no `tsconfig.json` to write, and no build
```
bartlomieju: `tsconfig.json` is optional might be better (this one gives the vibe that it's no possible to use tsconfig.json which si not true.

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
 
-:::note
+```sh
+deno run main.ts
+```
 
-**Deno type checks TypeScript in `strict mode` by default**, the TypeScript core
-team
-[recommends strict mode as a sensible default](https://www.typescriptlang.org/play/?#example/new-compiler-defaults).
+Deno also ships the type checker: `deno check` plays the role of `tsc --noEmit`,
```
bartlomieju, line 32: Mention LSP as well?

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+Deno treats executing TypeScript and type-checking it as two separate concerns:
+
+- **Execution**: when you `deno run` a TypeScript file, Deno strips the types
+  and hands the resulting JavaScript to V8. This is fast, happens in-process,
+  and is cached, but it does not look at whether the types are correct.
+- **Type checking**: `deno check` (or `deno run --check`) runs the real
+  TypeScript compiler over your code without executing it.
+
+This means **type errors do not stop your code from running unless you ask**.
+Checking a whole module graph is expensive, so Deno keeps it out of the hot
```
bartlomieju: Type-checking a whole module graph...

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+edit-run loop and lets you invoke it where it earns its cost: in your editor
+(via the Deno language server), in CI, and in `deno test`, which type-checks by
+default.
+
+A consequence worth knowing: Deno never emits JavaScript files. There is no
+`outDir`, no `dist/` directory, and no source map configuration — error stack
+traces point at your `.ts` sources directly. If you genuinely need `.js` output
```
bartlomieju: Maybe rephrase it - it's not true that Deno doesn't emit - it does emit into a global internal cache (but it's still inspectable). Maybe give a little info how to see the actually emitted file using `deno info`. Good callout to `deno transpile` and `deno pack` 👍

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+replacement:
+
+| In your Node.js project                         | In Deno                                                                                                         |
+| ----------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
+| `tsx`, `ts-node`, or `node` with type stripping | `deno run main.ts`                                                                                              |
+| `tsc --noEmit`                                  | `deno check`                                                                                                    |
+| `tsconfig.json`                                 | Optional. Auto-detected if present; Deno-first projects use `compilerOptions` in `deno.json`                    |
+| `typescript` as a devDependency                 | Nothing to install — the checker ships inside the `deno` binary                                                 |
+| `@types/node`                                   | Built in — `node:` imports are typed out of the box                                                             |
```
bartlomieju: Remove em-dashes

bartlomieju: This one needs to be more concrete - we have a default `@types/node` assigned in the runtime, but it's still possible to overwrite the version as users see fit. This should be expanded and preferably link to a section that goes into it in depth

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+| ESLint + Prettier with TypeScript plugins       | [`deno lint`](/runtime/reference/cli/lint/) and [`deno fmt`](/runtime/reference/cli/fmt/) handle `.ts` natively |
+
+Three differences trip up Node.js developers most often:
+
+**Imports use real file extensions.** In Deno you write
+`import { greet } from "./greet.ts"` — the extension is required, and it is the
+actual extension of the file on disk. If you are used to `tsc` with Node-style
+ESM, this replaces the counterintuitive habit of writing `./greet.js` to import
+a file named `greet.ts`. See [Modules](/runtime/fundamentals/modules/) for how
+resolution works, including npm packages.
```
bartlomieju, line 86: I think TS now has a setting for this - find it and link to it, add some more explanation.

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+
+**Deno runs the full TypeScript language.** Node's built-in type stripping only
+handles syntax it can erase; features that generate runtime code — enums,
+namespaces with runtime values, and parameter properties — require Node's
+experimental `--experimental-transform-types` flag. Deno runs all of them with
+no flag.
```
bartlomieju: Verify if this is correct (I think Node might have removed that flag in latest version?)

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+
+**Most of your `tsconfig.json` is about emit, and emit doesn't exist.** Options
```
bartlomieju: See previous comment - rephrase it so it's not confusing

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
-deno check
+# Type-check JavaScript files too, without adding @ts-check to each one
+deno check --check-js main.js
+```
 
-# Check a specific TypeScript file
-deno check module.ts
+`deno check` exits non-zero on errors, so it works as-is in CI. It can also
+check code blocks in JSDoc comments and markdown files — see
+[documentation tests](/runtime/test/doc_tests/).
```
bartlomieju: remove em-dash

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+## Configuring TypeScript compiler options
 
-```shell
-deno run --check module.ts
-# or also type check remote modules and npm packages
-deno run --check=all module.ts
+Deno's defaults are strict and modern, so most projects need no configuration at
+all. When you do want to change the checker's behavior, use the
+`compilerOptions` field in [`deno.json`](/runtime/fundamentals/configuration/):
+
```
bartlomieju: Mention how to use regular `tsconfig.json` (maybe even prefer that in this page)

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
 
-In order to avoid this, you will either need to:
+### Using an existing tsconfig.json {#tsconfig}
 
-- resolve the issue
-- use the `// @ts-ignore` or `// @ts-expect-error` pragmas to ignore the error
-- or skip type checking all together.
+You do not have to translate your configuration to try Deno. Each workspace
+directory containing a `deno.json` or `package.json` is probed for a
+`tsconfig.json` — if one exists, Deno automatically uses it for type checking
```
bartlomieju: remove em-dash

runtime/fundamentals/typescript.md
```diff
@@ -12,89 +12,203 @@ oldUrl:
...
+For Deno-first projects, prefer `compilerOptions` in `deno.json` over a separate
+`tsconfig.json`. See
+[Configuring TypeScript](/runtime/reference/ts_config_migration/) for supported
+fields, precedence rules, and project-reference behavior.
 
-```ts title="main.js"
+## Type checking JavaScript
+
+Deno runs JavaScript and TypeScript side by side, but only type-checks
+TypeScript files by default. To include a JavaScript file in checking, add a
```
bartlomieju: Slightly confusing because we mention `--check-js` flag before. Make sure it's mentioned here too and the previous occurrence links to this section

PR #3254, opened 2026-06-13
docs: add a Vite tutorial

bartlomieju:
A self-contained front-end tooling tutorial from the backlog, using a
stack that runs on Deno with no separate Node.js install.

The Vite tutorial scaffolds a project with the standard create-vite
templates, installs dependencies with deno install, runs the
hot-reloading dev server, and builds the production bundle, showing that
Deno reads package.json scripts as tasks. It is framework-agnostic: the
same flow works for the vanilla, React, Vue, and Svelte templates.

Every command was run and its output captured rather than written by
hand: scaffolding the vanilla-ts template, deno install, deno task dev
(hot reload), and deno task build into dist. The tutorial is filed under
Web frameworks and libraries alongside the existing Astro tutorial.

This PR originally also added a Lume tutorial; it was dropped because
Lume is Deno-native (no compatibility story to demonstrate) and
lume.land already documents it canonically, so a tutorial here would be
a thinner duplicate that diverges over time.

Build is clean and the full test suite passes.


examples/tutorials/lume.md
```diff
@@ -0,0 +1,100 @@
...
+title: "Build a static site with Lume"
+description: "Step-by-step tutorial on building a static site with Lume and Deno. Learn how to scaffold a project, add content pages, use layouts, run the dev server, and build for production."
+url: /examples/lume_tutorial/
+---
+
+[Lume](https://lume.land/) is a static site generator for Deno, inspired by
+Jekyll and Eleventy. It is small, fast, and unopinionated: you bring Markdown,
+JSX, Nunjucks, or any of a dozen supported formats, and Lume turns them into a
+static site. These very docs are built with Lume.
+
```
bartlomieju: Too marketing-like.

PR #3258, opened 2026-06-13
docs: content sweep wave A — fix stale facts in reference pages

bartlomieju:
First wave of a content sweep over runtime pages that had not been reviewed
in the recent docs work. These are small, high-confidence fixes, each
re-verified against Deno 2.8 before changing.

node_apis.md had three duplicate rows in the hand-maintained Globals table
(CustomEvent, structuredClone, URLSearchParams) and two dead anchors on
__dirname/__filename pointing at a section that does not exist on the page;
those now link to the Node compatibility page. wasm.md's "Imports" example
described importing ./values.js/getValue while the code imported
./time.ts/getTimeInSeconds, and its output block contained a stray V:\scratch
paste artifact and a duplicated run; both fixed. docker.md used deno test
--allow-none, which no longer exists (it is --permit-no-files now), and
referenced a denoland/deno:1.x image tag; updated. continuous_integration.md
pinned the canary matrix to v1.x while the basic pipeline above it already
used v2.x. stability_and_releases.md described LTS as a current channel, but
LTS maintenance ended on April 30 2026 with no further releases, so the lts
channel and forward-looking LTS language are removed in favor of a short
historical note.


runtime/fundamentals/stability_and_releases.md
bartlomieju, line 1: let's revert this page for now

runtime/reference/node_apis.md
bartlomieju, line 1: I feel like this whole list is not useful at all...

PR #3262, opened 2026-06-14
docs: content and style pass on the Security and permissions page

bartlomieju:
The Permissions section was the thin spot on the security page: two abstract
paragraphs that pointed at the reference without ever showing the model in
action, and missing a few facts that are part of the security model rather
than flag trivia. This expands it with a concrete before/after example
(default-deny, then a scoped --allow-net grant), the interactive prompt
flow, deny-flag precedence, and a note that -A turns the sandbox off
entirely.

It also fills two genuine gaps. A new subsection warns that --allow-run and
--allow-ffi escape the sandbox, since subprocesses and native code run
unsandboxed, so granting either is effectively --allow-all. Another covers
adjusting permissions at runtime through Deno.permissions, framed as dropping
access after startup to shrink the attack surface. The untrusted-code toolkit
now mentions Web Workers with a reduced permission set as Deno's own
in-runtime isolation. Along the way it fixes a Title-Cased heading, an
em-dash, and a curly apostrophe for house-style consistency. Flag-level
detail stays in the Permissions reference, which the new prose links to.

### DIFF
--- a/runtime/fundamentals/security.md
+++ b/runtime/fundamentals/security.md
@@ -109,14 +109,20 @@ declaring them in your configuration file.
 A few permissions grant access to things Deno cannot sandbox, so granting them
 is effectively granting full access:
 
-- `--allow-run` lets the program spawn subprocesses, and a subprocess runs with
-  its own permissions, not the restricted set of the Deno process that started
-  it. A program that can run `deno` or a shell can launch a second program with
-  full access. Limit it to specific executables where you can
-  (`--allow-run=git`), and never combine `--allow-run=deno` with a sandboxed
-  parent.
+- `--allow-run` lets the program spawn subprocesses. A subprocess runs as a
+  separate program with its own permissions, not the restricted set you granted
+  the Deno process, so whatever it does happens outside the sandbox. This is why
+  `--allow-run=deno` is especially dangerous: a script that can start a new
+  `deno` process can start it with `--allow-all`, inheriting none of the
+  parent's limits and escaping the sandbox entirely. Grant `--allow-run` only
+  for specific, trusted executables where you can (`--allow-run=git`), and avoid
+  giving a sandboxed program the ability to launch `deno` or a shell.
 - `--allow-ffi` loads native libraries through [FFI](/runtime/fundamentals/ffi/)
-  or Node-API addons. Native code is not bound by Deno's permissions at all.
+  or Node-API addons. Deno enforces permissions in the JavaScript layer, but a
+  native library runs as compiled machine code in the same process and can issue
+  system calls directly. Once loaded, it can read files, open sockets, or do
+  anything the operating system lets the process do, regardless of which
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/security.md
```diff
@@ -62,15 +62,84 @@ underlying operating system.
...
+
+A few permissions grant access to things Deno cannot sandbox, so granting them
+is effectively granting full access:
+
+- `--allow-run` lets the program spawn subprocesses, and a subprocess runs with
+  its own permissions, not the restricted set of the Deno process that started
+  it. A program that can run `deno` or a shell can launch a second program with
+  full access. Limit it to specific executables where you can
+  (`--allow-run=git`), and never combine `--allow-run=deno` with a sandboxed
+  parent.
```
bartlomieju: "and never combine `--allow-run=deno` with a sandboxed parent." needs better explanation

runtime/fundamentals/security.md
```diff
@@ -62,15 +62,84 @@ underlying operating system.
...
+- `--allow-ffi` loads native libraries through [FFI](/runtime/fundamentals/ffi/)
+  or Node-API addons. Native code is not bound by Deno's permissions at all.
```
bartlomieju: Give some more explanation here

runtime/fundamentals/security.md
```diff
@@ -62,15 +62,84 @@ underlying operating system.
...
+
+Treat both as equivalent to `--allow-all` when deciding whether to trust the
+code you are running.
+
+### Adjusting permissions at runtime
+
+A program can inspect and tighten its own permissions through the
+[`Deno.permissions`](/api/deno/#permissions) API. `query` reports whether a
+permission is granted, `request` prompts for one on demand, and `revoke`
+downgrades a granted permission back to the prompt state:
```
bartlomieju: Use full API names (`Deno.permissions.query`) and link them to proper pages/anchors

runtime/fundamentals/security.md
bartlomieju: Expanded in bff85fa5: spelled out that a subprocess runs with its own permissions outside the sandbox, and why `--allow-run=deno` specifically defeats it (the child can be started with `--allow-all` and inherit none of the parent's limits).

runtime/fundamentals/security.md
bartlomieju: Expanded in bff85fa5: explained that Deno enforces permissions in the JS layer, but a native library runs as machine code in the same process and can issue syscalls directly, so it can do anything the OS process can regardless of `--allow-*` flags.

runtime/fundamentals/security.md
bartlomieju: Done in bff85fa5: now uses full `Deno.permissions.query`/`request`/`revoke` names, each linked to its API anchor (#query-permissions, #request-permissions, #revoke-permissions).

PR #3263, opened 2026-06-14
docs: content and prose pass on the Modules page

bartlomieju:
The Modules page redirects the old module_metadata URL but no longer
documented import.meta anywhere, leaving that content only in the API
reference. This adds an Import metadata section covering url, main, filename,
dirname, and resolve, including the common import.meta.main pattern for
making a file work both as an entry point and as a library, and links the
module metadata tutorial for a full walkthrough.

It also adds a Dynamic imports section. import() was already used in the
data URL examples but never introduced, so the new section explains
runtime-computed specifiers, conditional loading, and how dynamic imports
interact with permissions (literal specifiers join the static graph and need
no flags, computed ones are checked at runtime). Finally it tightens the
marketing-heavy intro into a single concrete paragraph that notes CommonJS is
supported for Node compatibility, and replaces an em-dash in a code string
for house-style consistency.

### DIFF
--- a/runtime/fundamentals/modules.md
+++ b/runtime/fundamentals/modules.md
@@ -11,15 +11,13 @@ oldUrl:
 
 Deno uses
 [ECMAScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)
-as its default module system to align with modern JavaScript standards and to
-promote a more efficient and consistent development experience. It's the
-official standard for JavaScript modules, allows for better tree-shaking,
-improved tooling integration, and native support across different environments.
+as its module system, the official standard for JavaScript. You share code
+between files with standard `import` and `export` statements and run it
+directly, with no bundler or build step. This is the same module system browsers
+use, so the code you write stays portable across environments.
 
-By adopting ECMAScript modules, Deno ensures compatibility with the
-ever-evolving JavaScript ecosystem. For developers, this means a streamlined and
-predictable module system that avoids the complexities associated with legacy
-module formats like CommonJS.
+Deno does not use CommonJS (`require`) by default, but
+[supports it for Node compatibility](/runtime/fundamentals/node/).
 
 ## Importing modules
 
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/modules.md
```diff
@@ -11,15 +11,13 @@ oldUrl:
...
+between files with standard `import` and `export` statements and run it
+directly, with no bundler or build step. This is the same module system browsers
+use, so the code you write stays portable across environments.
 
-By adopting ECMAScript modules, Deno ensures compatibility with the
-ever-evolving JavaScript ecosystem. For developers, this means a streamlined and
-predictable module system that avoids the complexities associated with legacy
-module formats like CommonJS.
+Deno does not use CommonJS (`require`) by default, but
+[supports it for Node compatibility](/runtime/fundamentals/node/).
```
bartlomieju: I think this is misleading. What does default mean in this context. I'm sort of thinking that this page's intro needs a concrete overhaul to mention both system are equally supported, ESM is preferred, there are some limitation (or features) on how CommonJS functions.

bartlomieju: Reworked the intro in 97242720. Dropped the misleading "does not use CommonJS by default" line. It now states that Deno fully supports both ESM and CommonJS and they interoperate, with ESM recommended for new code, and spells out how Deno picks the module type: `.mjs` is always ESM, `.cjs` is always CommonJS, and `.js`/`.ts` default to ESM unless a `package.json` sets `"type": "commonjs"`. Links to the CommonJS support section for require/resolution/interop details.

runtime/fundamentals/modules.md
```diff
@@ -11,15 +11,23 @@ oldUrl:
...
-module formats like CommonJS.
+as its primary module system, the official standard for JavaScript. You share
+code between files with standard `import` and `export` statements and run it
+directly, with no bundler or build step. This is the same module system browsers
+use, so the code you write stays portable across environments.
+
+Deno also runs Node's older CommonJS modules (`require` and `module.exports`),
+so existing npm packages and Node projects work without conversion. The two
+systems are fully supported and interoperate, but ECMAScript modules are
+recommended for new code; reach for CommonJS mainly when a dependency or an
```
bartlomieju, line 22: Should mention that this works in user code too...

PR #3264, opened 2026-06-14
docs: add a deployment hub page for the runtime section

bartlomieju:
The docs cover plenty of ways to run a Deno app in production, but the
guidance is scattered with nothing tying it together. The Docker
reference lives under Runtime, the platform walkthroughs (AWS Lambda,
Lightsail, Google Cloud Run, DigitalOcean, Kinsta, Cloudflare Workers)
live as examples, and the Deno Deploy section covers only the managed
platform. A reader asking "I have an app, where do I run it?" had no
single page to start from.

This adds a Deploying your app page to the runtime Guides sidebar. It
frames the choice between a managed platform and self-hosting, then links
out to Deno Deploy, the Docker reference, the per-provider tutorials, and
the deno compile path for a standalone binary, plus a short production
checklist covering permissions, environment variables, observability,
and CI. It is a hub of links rather than new how-to content, so nothing
moves and no URLs change. It sits under /runtime/ to stay clearly
separate from the managed Deno Deploy section in the top nav.

### DIFF
--- a/runtime/deploy/index.md
+++ b/runtime/deploy/index.md
@@ -1,0 +1,85 @@
+---
+title: "Deploying your app"
+description: "Ways to run a Deno app in production: the managed Deno Deploy platform, containers and Docker, cloud and serverless providers, and self-hosting a standalone binary."
+---
+
+Once your app runs locally, you have a few ways to get it into production.
+Broadly, you either let a managed platform run it for you, or you run it
+yourself on infrastructure you control. This page maps the options and points to
+the guide for each.
+
+## Choosing an option
+
+- **Want the least operational overhead?** Use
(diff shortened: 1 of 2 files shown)


runtime/deploy/index.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Deploying your app"
+description: "Ways to run a Deno app in production: the managed Deno Deploy platform, containers and Docker, cloud and serverless providers, and self-hosting a standalone binary."
```
bartlomieju, line 4: Missing last_updated

bartlomieju: Added `last_modified: 2026-06-14` to the frontmatter in 612d61b9 (the field is named `last_modified` to match the other pages).

PR #3265, opened 2026-06-14
docs: expand the Config files page with an example and package.json story

bartlomieju:
The Config files page was almost pure navigation: prose plus a list of links
into the deno.json reference, without ever showing what a deno.json actually
looks like. This adds a minimal example up front, states that a configuration
file is optional, explains that config is discovered up the directory tree so
a project's settings apply to every file under it, and links the workspaces
page for monorepos.

It also leans further into package.json. The section now frames the
deno.json-and-package.json dual model directly: Deno reads dependencies and
scripts from package.json while its own tooling settings come only from
deno.json, and the two can coexist. That sets up the incremental adoption
story, so the section links the Migrate from Node.js guide for the
walkthrough and the Node compatibility page for how the runtime maps Node's
APIs and module resolution. The stale node-compatibility anchor is dropped in
favor of a clean link to the page.

### DIFF
--- a/runtime/fundamentals/configuration.md
+++ b/runtime/fundamentals/configuration.md
@@ -1,18 +1,51 @@
 ---
-last_modified: 2026-03-09
+last_modified: 2026-06-14
 title: "Config files"
-description: "How Deno projects are configured: the deno.json file, .jsonc support and discovery, how package.json is used for Node compatibility, and an overview of what you can configure. See the deno.json reference for every field."
+description: "How Deno projects are configured: first-class package.json support, the deno.json file for Deno's own tooling, .jsonc support and discovery, and an overview of what you can configure. See the deno.json reference for every field."
 oldUrl:
   - /runtime/manual/getting_started/configuration_file/
 ---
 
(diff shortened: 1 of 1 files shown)


bartlomieju (commented):
I think we should rewrite this page to focus more on package.json - there's no mention that you can use package.json as first class and don't need deno.json unless you want to configure tools like fmt...

Also last modified field is not updated

bartlomieju:
Refocused the page on package.json in 2669b727:

- The intro now leads with the two-file model: `package.json` and `deno.json` are both first-class and both optional, followed by a rule of thumb — use `package.json` for dependencies and scripts (no `deno.json` needed), add `deno.json` only to configure Deno's own tooling (fmt, lint, TS compiler, tasks).
- A dedicated **package.json** section now comes first, leading with first-class support and a `deno install` / `deno task` example, and stating plainly that most Node projects run with no `deno.json` and no conversion.
- The **deno.json** section follows, framed as where you configure Deno itself.
- Also updated the page description and bumped `last_modified` to 2026-06-14.

PR #3281, opened 2026-06-15
docs: show difficulty on example cards

bartlomieju:
Surface each example's difficulty (beginner, intermediate, or advanced,
from the script's existing @difficulty tag) as a small colored dot and
label on the landing-page cards, so the relative level is scannable at a
glance. It flows through the same generator path as the card
descriptions; only examples declare a difficulty, so tutorials and videos
stay unbadged.

Split out from the original combined PR since the badge is more
debatable than the other changes. Stacked on #3278 (the Start
here/cookbook PR) to avoid touching the same landing-page code twice;
GitHub will retarget this to main once #3278 merges. Easy to close on its
own if the badge isn't wanted.

### DIFF
--- a/_components/LandingPage.tsx
+++ b/_components/LandingPage.tsx
@@ -27,7 +27,10 @@ function FeaturedStar() {
 }
 
 export default function LandingPage(
-  { descriptions }: { descriptions?: Record<string, string> },
+  { descriptions, difficulties }: {
+    descriptions?: Record<string, string>;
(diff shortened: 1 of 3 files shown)


bartlomieju:
Closing this one. Difficulty is a subjective signal and would only appear on example cards (not tutorials/videos), making the grid inconsistent, and the landing page is already signal-heavy. Keeping the @difficulty metadata in place as-is for potential future use; just not surfacing it on the cards.

PR #3285, opened 2026-06-16
docs: add migrate-from npm, pnpm, and Yarn pages

bartlomieju:
The migrate section is meant to cover moving to Deno from anywhere, and
people switching from a specific package manager search for that tool by
name. Until now that audience was served by a single unified guide,
switch_package_manager, which is one URL competing for "migrate from npm",
"migrate from pnpm", and "migrate from yarn" at once. This splits it into a
dedicated page per package manager so each one can own its search term and
its tool-specific advice.

Each page leads with the same honest framing, that a package manager is
configuration rather than code, so Deno reads the existing package.json and
most projects run unchanged, then carries the command cheatsheet and the
parts specific to that tool. npm and Yarn workspaces live in package.json
and are read directly, so there is nothing to convert; the pnpm page
documents converting pnpm-workspace.yaml (packages and catalogs) into
deno.json by hand. The Yarn page covers Plug'n'Play, which Deno does not
implement. The npm page also carries the incremental-adoption framing (use
Deno as your package manager while still running the app with Node) and
absorbs the redirect from the old unified URL.

The previous switch_package_manager guide is removed and its URL redirects
to the npm page; its fuller command mapping (audit, why, ci, dev
dependencies) and the node_modules layout note are folded into the per-tool
pages, and the sidebar now lists the three pages in its place. The pages are
written against currently shipped behavior only, so they stay correct
regardless of in-flight runtime changes, and every command and config field
was verified against the CLI reference or by running Deno.


bartlomieju:
Closing in favor of a single combined PR that targets the `2.9` branch and adds the npm/yarn/pnpm/Bun migration pages together. Deno 2.9 seeds `deno.lock` from `bun.lock` (denoland/deno#35394), which this page predates. The corrected Bun page is carried into the new PR.

bartlomieju:
Superseded by #3353 (targets the `2.9` branch).

Issue #2708, opened 2025-11-05
Feedback: /runtime/fundamentals/security/ - Needs Improvement


denobot:
## Documentation Feedback

**Path:** /runtime/fundamentals/security/
**Helpful:** No ❌

**Feedback:**
> Regarding the permissions for the `--allow-sys` flag, the documentation also needs to be fully updated to cover the APIs utilized by `node:os`.

---
*This issue was automatically created from the documentation feedback system.*

PR #3291, opened 2026-06-17
docs: clarify permissions and lint configuration

### LINKED ISSUE #1642: Add information about `NotCapable` error to permission system overview
When a user encounters a `NotCapable` error, that means Deno's permission system denied access.

It would be good to outline this on https://docs.deno.com/runtime/fundamentals/security

Ref: https://github.com/denoland/deno/issues/28947

### LINKED ISSUE #2434: Feedback: /lint/ - Needs Improvement
## Documentation Feedback

**Path:** /lint/
**Helpful:** No ❌

**Feedback:**
> "You can enable sets of rules in deno.json(c) by adding their tags (e.g. recommended, react) to the lint.rules.tags array." ok but what about the rules with no tag lol

---

### LINKED ISSUE #2692: Feedback: /lint/rules/eqeqeq/ - Needs Improvement
## Documentation Feedback

**Path:** /lint/rules/eqeqeq/
**Helpful:** No ❌

**Feedback:**
> I'm missing a way to configure an exception to this rule. I would like to be able to write : `if (a != null) {}`

**GitHub User:** @edouardmisset

---

### LINKED ISSUE #2846: Document limitations of `--allow-run=...` once using `--allow-write`
When someone has `--allow-write`, it's possible to just overwrite the location of an allowed binary:

```ts
#!/usr/bin/env -S deno run --allow-write --allow-run=/tmp/binary
Deno.writeFileSync("/tmp/binary", NEW_BINARY);
new Deno.Command("/tmp/binary").output();
```

Additionally, symlink creation requires `--allow-write` and so someone could modify symlinks to point wherever (thought they could just write whatever bytes to whatever binary at that point). If someone wants to prevent execution of arbitrary user written subprocesses, they must not allow writes to any file that is in the allow-run list, or that any file in the allow-run list symlinks to.

### LINKED ISSUE #2890: Feedback: /lint/rules/ban-unused-ignore/ - Needs Improvement
## Documentation Feedback

**Path:** [/lint/rules/ban-unused-ignore/](https://docs.deno.com/lint/rules/ban-unused-ignore/)
**Helpful:** No ❌

**Feedback:**
> unclear examples of valid and invalid

**GitHub User:** @QurioSapien

---

bartlomieju:
A batch of small clarifications from feedback issues, each verified against
Deno 2.8:

- The security guide now documents the `NotCapable` error thrown when access
  is denied, with an example of catching it, and warns that granting
  `--allow-write` over a directory that holds an `--allow-run` binary lets a
  program overwrite that binary and escape the sandbox.
- The `eqeqeq` rule page notes it has no options and shows how to allow an
  intentional `!= null` with an ignore directive.
- The `ban-unused-ignore` page has clearer necessary-vs-superfluous examples.
- The lint docs explain that untagged rules must be enabled individually via
  `include`.

Closes #1642, closes #2846, closes #2692, closes #2890, closes #2434.

Note: the node:os to --allow-sys mapping table was removed per review, so
#2708 is left open.

### DIFF
--- a/runtime/fundamentals/security.md
+++ b/runtime/fundamentals/security.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-06-14
+last_modified: 2026-06-17
 title: "Security and permissions"
 description: "A guide to Deno's security model: secure-by-default execution, the permission sandbox, evaluating and executing untrusted code, and the permission broker. See the Permissions reference for the flags."
 ---
@@ -84,6 +84,20 @@ reads to one directory, `--allow-env=API_KEY` to a single variable, and so on. A
 bare `--allow-net` with no value grants the whole category, which is broader
 than most programs need.
 
+When an operation is refused (because the permission was not granted or was
+explicitly denied), Deno throws a `NotCapable` error. You can catch it like any
+other error if you want to handle missing permissions gracefully:
+
+```ts
+try {
+  await Deno.readTextFile("/etc/hosts");
+} catch (err) {
+  if (err instanceof Deno.errors.NotCapable) {
+    console.error("Missing read permission, run again with --allow-read");
+  }
+}
+```
+
 When stdout is a terminal and you have not passed a flag, Deno pauses and asks
 instead of failing, so you can grant access interactively as it is requested:
 
@@ -123,6 +137,13 @@ is effectively granting full access:
   system calls directly. Once loaded, it can read files, open sockets, or do
   anything the operating system lets the process do, regardless of which
   `--allow-*` flags you passed.
+
+Be careful when combining `--allow-write` with `--allow-run`. Write access to a
+directory that contains an allowed executable (or to the executable itself) lets
+a program overwrite that binary, or replace it with a symlink, so the next
+subprocess it spawns runs attacker-controlled code with the host's privileges.
+Avoid granting write access to directories that hold binaries you allow with
+`--allow-run`.
 
 Treat both as equivalent to `--allow-all` when deciding whether to trust the
 code you are running.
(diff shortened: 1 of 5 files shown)


runtime/fundamentals/security.md
```diff
@@ -124,6 +138,13 @@ is effectively granting full access:
   anything the operating system lets the process do, regardless of which
   `--allow-*` flags you passed.
 
+Be careful when combining `--allow-write` with `--allow-run`. Write access to a
+directory that contains an allowed executable (or to the executable itself) lets
+a program overwrite that binary, or replace it with a symlink, so the next
```
bartlomieju: This is not true, symlinks require full permissions, please fix it

runtime/reference/permissions.md
```diff
@@ -382,6 +382,21 @@ deno run --allow-sys --deny-sys="networkInterfaces" script.ts
 deno run --deny-sys script.ts
 ```
 
+Node compatibility APIs route through the same permission. Functions from
+[`node:os`](https://nodejs.org/api/os.html) map to these `--allow-sys` API
+names:
+
```
bartlomieju: This seems stray, remove

runtime/fundamentals/security.md
bartlomieju: Fixed in 71a9e68a: removed the symlink claim and kept the binary-overwrite warning.

runtime/reference/permissions.md
bartlomieju: Removed the node:os mapping table in 71a9e68a. Left #2708 open.

Issue #1, opened 2023-08-15
Update manual and deploy docs content for 1.36


kwhinnery:
Bring in new content from latest feature release from the Manual.

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

denobot:
Additional feedback:

> NIGGER

**GitHub User:** @t3gg @t3gg @t3gg @t3gg

PR #3294, opened 2026-06-17
docs: prose fixes on Desktop apps pages

bartlomieju:
A prose pass over the new Desktop apps section surfaced three small
correctness issues. The section overview claimed in-process bindings mean
"no serialization step" between Deno and the webview, which contradicts the
Bindings and Comparison pages that both state values are still encoded as
they cross the call boundary; what bindings actually avoid is the
cross-process round-trip, so the overview now says that. Two snippets on the
HTTP serving page read DENO_SERVE_ADDRESS without a non-null assertion and
would not type-check, unlike the matching code on the Windows page. The
Notifications icon example called encodeBase64 with no import.

These are the only correctness issues found; the rest of the section reads
well. A few things I was unsure about are left for review rather than changed:
the distribution CI matrix uses a macos-15-intel runner label I could not
confirm exists, the comparison table lists the process model as
"Multi-thread" while the Bindings page describes WebView as a coordinated
process group, and the section index does not link to the deno desktop CLI
reference page.

### DIFF
--- a/runtime/desktop/bindings.md
+++ b/runtime/desktop/bindings.md
@@ -1,7 +1,7 @@
 ---
 last_modified: 2026-06-16
 title: "Bindings"
-description: "Call Deno-side functions from webview JavaScript via win.bind(): type-safe RPC over in-process channels, with no IPC and no serialization beyond the call boundary."
+description: "Call Deno-side functions from webview JavaScript via win.bind(): type-safe RPC over in-process channels, encoded at the boundary with no cross-process round-trip."
 ---
 
 :::info Coming in Deno 2.9
--- a/runtime/desktop/bindings.md
+++ b/runtime/desktop/bindings.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-06-16
+last_modified: 2026-06-18
 title: "Bindings"
 description: "Call Deno-side functions from webview JavaScript via win.bind(): type-safe RPC over in-process channels, encoded at the boundary with no cross-process round-trip."
 ---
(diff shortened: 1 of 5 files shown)


bartlomieju:
Open questions left for review (not changed in this PR):

1. **`distribution.md` — `macos-15-intel` runner label.** The CI matrix uses
   `{ os: macos-15-intel, target: x86_64-apple-darwin }`. I couldn't confirm
   that's a valid GitHub-hosted runner label (the established ones are
   `macos-13` for Intel, `macos-14`/`macos-15` for Apple silicon). If it's
   wrong, anyone copy-pasting the workflow gets a failing job. Is
   `macos-15-intel` real, or should it be `macos-13`?

2. **`comparison.md` — "Process model: Multi-thread".** The at-a-glance table
   lists Multi-thread for `deno desktop`, but `bindings.md` describes the
   runtime and backend as same-address-space threads *only for CEF* and
   WebView as a "coordinated process group" (i.e. multi-process). Is WebView
   actually multi-process? If so, should the table say "Multi-thread (CEF)"?

3. **`bindings.md` description frontmatter — "no serialization beyond the call
   boundary".** Defensible, but reads as nearly contradicting the body ("still
   encoded as JSON"). It's the same framing that made the index page
   inaccurate (fixed here). Keep as-is, or reword to "encoded at the boundary,
   no cross-process round-trip"?

4. **`index.md` "What's in this section" list doesn't link to the CLI
   reference** (`runtime/reference/cli/desktop.md`), even though that page
   links back into the section. Should the section index link to the
   `deno desktop` CLI reference? (Easy add if yes.)

crowlKats:
`macos-15-intel` is real.

bartlomieju:
All four open questions are resolved: #1 `macos-15-intel` confirmed real (no change), and @crowlKats's review edits handled #2 (process model now `Multi-thread (CEF) / process group (WebView)`), #3 (bindings description reworded), and #4 (added the `deno desktop` CLI reference link). I bumped `last_modified` on `bindings.md`/`comparison.md` so the freshness check passes. Should be ready to land once CI is green.

PR #3296, opened 2026-06-18
docs: apply npm @types to every import without per-line @ts-types

### LINKED ISSUE #1713: Can we define types globally on a project?
Docs feedback for `/runtime/fundamentals/typescript/`

> For an NPM package that doesn't offer native typings: Is there a way to set the types to use globally so one does not need to add  
`// @ts-types="npm:@types/somepackage"`  on every import? It seems unintuitive having to declare it each time.

bartlomieju:
A reader on the TypeScript page asked whether npm package types can be set
once for a project instead of repeating @ts-types=\"npm:@types/...\" on every
import. There is no deno.json setting that maps a bare npm specifier to its
@types package globally, so the page now documents the idiomatic alternative:
re-export the package from a small typed wrapper module that carries the
@ts-types annotation, then import the wrapper everywhere.

I verified the pattern locally with deno check: importing the wrapper applies
the real @types/lodash types at every call site, and a wrong-typed use fails
type checking as expected.

Closes #1713

### DIFF
--- a/runtime/fundamentals/typescript.md
+++ b/runtime/fundamentals/typescript.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-06-12
+last_modified: 2026-06-18
 title: "TypeScript support"
 description: "TypeScript is a first-class language in Deno. Run .ts files directly with no build step, type-check with deno check, reuse your tsconfig.json, and skip the toolchain you needed under Node.js."
 oldUrl:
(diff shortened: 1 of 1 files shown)


bartlomieju:
Reworked. The wrapper-module approach was awkward — the idiomatic answer is to add the `@types` package as a dependency and let Deno resolve it automatically from a local `node_modules`, like `tsc`:

```sh
deno add npm:lodash npm:@types/lodash
```
```ts
import { capitalize } from "lodash"; // fully typed, no @ts-types
```

Verified with `deno check` that the `@types` apply to both bare and `npm:` imports when a local `node_modules` exists (`nodeModulesDir: "auto"` or a `package.json`), and that it does not apply under the default global cache — the page now notes that and keeps `@ts-types` as the fallback for that case.

runtime/fundamentals/typescript.md
```diff
@@ -341,6 +341,35 @@ the corresponding `@types` package:
...
+### Applying types to every import
+
+`@ts-types` annotates a single import, so for a package you use throughout a
+project, repeating it at every import site gets tedious. Instead, add the
+`@types` package as a dependency. When the types live in a local `node_modules`
+directory, Deno picks them up automatically for every import of the package, the
+same way `tsc` does, with no annotation:
+
+```sh
+deno add npm:lodash npm:@types/lodash
```
bartlomieju:
```suggestion
deno add lodash @types/lodash
```

PR #3342, opened 2026-06-24
Adding `deno.desktop` as a documented library

pedropalhari:
This PR is just to bring attention to the community. So `deno desktop` autocompletions such as `new Deno.BrowserWindow` get properly typechecked and are supported on LSP.

Probably missing a section for using it as:

```js
/// <reference lib="deno.desktop" />
```

Which I think could be very nice on the initial single file examples.


bartlomieju (approved):
Thanks

PR #3346, opened 2026-06-25
WebView backend is not the default

tredondo:
Reworded the description of the WebView backend for clarity.

### DIFF
--- a/runtime/desktop/index.md
+++ b/runtime/desktop/index.md
@@ -28,9 +28,9 @@ integration.
 
 `deno desktop` is opinionated about those tradeoffs:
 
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks, but WebView is the default backend: https://github.com/denoland/deno/pull/35442

tredondo:
@bartlomieju Oh, sorry, I misinterpreted [your comment](https://github.com/denoland/deno/issues/35433#issuecomment-4776377899) "Yeah, that was mistake in the doc" to mean the docs should be fixed reflect the default backend being CEF. All good now.

PR #3374, opened 2026-06-30
docs: expand CSS module import documentation

### LINKED ISSUE #3257: Document CSS module imports (`with { type: "css" }`) under --unstable-raw-imports
## Feature

CSS module imports landed in Deno via [denoland/deno#35093](https://github.com/denoland/deno/pull/35093) (behind the `--unstable-raw-imports` flag, towards denoland/deno#11961).

Importing a stylesheet with an import attribute now evaluates to a `CSSStyleSheet`, matching what Chrome and Firefox ship:

```ts
import sheet from "./styles.css" with { type: "css" };
console.log(sheet instanceof CSSStyleSheet);
for (const rule of sheet.cssRules) {
  console.log(rule.cssText);
}
```

Dynamic imports work too:

```ts
const { default: sheet } = await import("./styles.css", {
  with: { type: "css" },
});
```

The primary use case is running unmodified browser module graphs in Deno (SSR and testing of web components), where a CSS import currently kills module loading before any code runs.

## What needs documenting

- How to enable the feature: `--unstable-raw-imports` flag (and the `"unstable": ["raw-imports"]` field in `deno.json`).
- Static and dynamic `with { type: "css" }` import syntax.
- The minimal `CSSStyleSheet` surface that Deno implements:
  - `cssRules` — returns a **frozen array** of `CSSRule` (not a live `CSSRuleList`); each access creates a fresh array.
  - `replace(text)` / `replaceSync(text)` — top-level `@import` rules are dropped (matching the CSSOM constructed-stylesheet behavior).
  - `CSSRule.cssText` — verbatim text of one top-level rule.
  - `new CSSStyleSheet()` constructor (constructor `options` like `media`/`disabled`/`baseURL` are not supported).
- Important limitations to call out:
  - Deno has no DOM, so a sheet can't be adopted anywhere; the implementation is backed by the raw CSS text.
  - `cssRules` uses a naive top-level rule split, **not** a real CSS parser/object model — methods like `insertRule`/`deleteRule` are not implemented.
  - Reading a CSS file requires read permission (`--allow-read`).
- Note the unstable status (API yet to be vetted; subject to change).

## References

- Implementation PR: denoland/deno#35093
- Tracking issue: denoland/deno#11961
- Spec: https://html.spec.whatwg.org/multipage/webappapis.html#css-module-script

bartlomieju:
Expands the `with { type: "css" }` documentation in the modules guide to
cover the details requested in #3257, which the existing short section
left out.

Adds the dynamic `import(..., { with: { type: "css" } })` form, the
read-permission requirement for loading a stylesheet from disk, and a
description of the slice of the `CSSStyleSheet` interface that Deno
actually implements: `cssRules` as a frozen array of `CSSRule` (not a
live `CSSRuleList`), `CSSRule.cssText`, `replace`/`replaceSync` dropping
top-level `@import` rules, and the `new CSSStyleSheet()` constructor
without support for its `options` argument. Also spells out the
limitations that follow from Deno having no DOM: a sheet can't be adopted
anywhere, the rule list comes from a naive top-level split, and
`insertRule`/`deleteRule` are not implemented.

I verified each of these claims against the implementation in
`ext/web/css_stylesheet.rs` before documenting them.

Closes #3257.

### DIFF
--- a/runtime/fundamentals/modules.md
+++ b/runtime/fundamentals/modules.md
@@ -1,5 +1,5 @@
 ---
-last_modified: 2026-06-25
+last_modified: 2026-06-30
 title: "Modules"
 description: "Learn how Deno's ECMAScript module system works: importing local and third-party modules, import attributes, import maps, and supported import types such as Wasm and data URLs."
 oldUrl:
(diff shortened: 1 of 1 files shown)


runtime/fundamentals/modules.md
```diff
@@ -183,6 +183,44 @@ console.log(sheet instanceof CSSStyleSheet);
...
+
+```ts
+const { default: sheet } = await import("./styles.css", {
+  with: { type: "css" },
+});
+```
+
+Reading a CSS file from disk needs read permission, so run with `--allow-read`
+(or the narrower `--allow-read=./styles.css`).
+
```
bartlomieju: This is true only if the specifier is non-analyzable (eg. `import(foo + "styles.css")`).

bartlomieju: Good catch, fixed. Reworded to say static imports and analyzable dynamic specifiers load as part of the module graph and need no permission, and that only a non-analyzable dynamic specifier (e.g. `import(base + "styles.css")`) requires `--allow-read`.

PR #3401, opened 2026-07-08
docs: clarify Deno.cron runtime behavior

### LINKED ISSUE #21555: `Deno.cron` Day-of-Week mismatch: Numeric weekdays off by one
**Version:** Deno 1.38.5

There is an inconsistency with the numeric representation of weekdays in the `Deno.cron` function, it is off by one compared to "standards". 

**Expected Behavior:**
Based on conventional system cron logic (as detailed in [Wikipedia's cron overview](https://en.wikipedia.org/wiki/Cron)), the expression `* * * * 3` should denote every minute of every Wednesday. This follows the typical numbering of days where Sunday is 0, Monday is 1, and so forth.

**Actual Behavior:**
Contrary to expectations, in Deno 1.38.5, the expression `* * * * 4` is required to trigger events every minute on Wednesdays. This deviation originates from the Rust crate `saffron`, which Deno utilizes for cron parsing. Saffron, as explained in a [Cloudflare blog post](https://blog.cloudflare.com/using-one-cron-parser-everywhere-with-rust-and-saffron/), is influenced by the Quartz scheduler's cron parser. Quartz adopts a different standard where the week starts with 1 (Sunday) and ends with 7 (Saturday).

**Concern:**
The main concern is the adoption of Quartz's unconventional weekday numbering in Deno, which contradicts the more widely accepted system cron standard (0 for Sunday, 1 for Monday, etc.). This discrepancy can lead to confusion and potential errors in cron job scheduling within the Deno environment.

**Alternative Solution:**
As a response to this and more issues with existing rust cron implementations, I have developed an alternative crate, [`croner-rust`](https://github.com/Hexagon/croner-rust). This crate aims to align more closely with the standard cron behavior, and extend them with commonly adopted additional features. A detailed comparison between `croner-rust` and the existing `cron` and `saffron` crates can be found at [Why Croner instead of Cron or Saffron?](https://github.com/hexagon/croner-rust#why-croner-instead-of-cron-or-saffron).

sanekb:
### Changes

1. Runtime (`runtime/fundamentals/cron.md`)

    * Expanded Cron Syntax section with clear reference table and detailed explanation of supported fields.
    * Added Advanced Features and Examples sections based on `cron.rs` analysis and test cases.

2. Deploy (`deploy/reference/cron.md`)

    * Added reference to the Runtime Docs to avoid information duplication and maintain a single source of truth.
    * Small link fixes.

### Testing

- [x] `deno fmt --check` of both files passes

Closes denoland/deno#21555

### DIFF
--- a/runtime/fundamentals/cron.md
+++ b/runtime/fundamentals/cron.md
@@ -44,7 +44,8 @@ Deno.cron("hourly-task", { hour: { every: 1 } }, () => {
 
 :::note
 
-All times are UTC — this avoids ambiguity around daylight saving transitions.
+All times are **UTC** — this avoids ambiguity around daylight saving
+transitions.
 
 :::
 
@@ -80,18 +81,15 @@ Each field accepts an exact value, `*` (every value), a range (`1-5`), a list
 
 :::note
 
-Unlike traditional UNIX cron engines (which use `0-6` where `0` is Sunday), Deno
-maps numeric days of the week from **1** to **7** (1 = Sunday, 7 = Saturday).
+Unlike traditional UNIX cron engines (which use `0–6`, where `0` is Sunday),
+Deno maps numeric days of the week as **`1–7`** (where `1` is Sunday and `7` is
+Saturday).
 
 :::
 
 ### Advanced Features
 
 Deno's parser introduces powerful features for advanced scheduling:
-
-- **Wraparound Ranges:** Inverted ranges that cross time boundaries are
-  supported. Useful for night-time windows or end-of-month + start-of-next-month
-  logic.
 
 - **Last Day of Month:** Specifying `L` in the _Day of Month_ field triggers the
   task on the exact last day of the current month (28, 29, 30, or 31).\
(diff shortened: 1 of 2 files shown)


runtime/fundamentals/cron.md
```diff
@@ -27,41 +27,101 @@ reference</a>
...
+  task on the exact last day of the current month (28, 29, 30, or 31).\
+  Specifying `L-<offset>` triggers on the last day minus the offset (e.g., `L-1`
+  is the second-to-last day). The offset can be from `1` to `30`.
+
+- **Nearest Weekday:** Specifying `<day>W` triggers on the closest weekday
+  (Mon–Fri) to the given date.\
+  `LW` or `L-<offset>W` works relative to the end of the month.
+
+- **Last Day of Week:** Specifying `<day>L` triggers on the last specific day of
+  the week in that month. Standalone `L` is equivalent to `7L`.
```
bartlomieju: This isn't what the runtime does. A bare `L` in the day-of-week field parses to `DaysOfWeek::Days(1u8 << 6)` — a plain match on **every Saturday** — not the last-Saturday special case:

```rust
// ext/cron/cron.rs
if input == "L" {
  return Ok(DaysOfWeek::Days(1u8 << 6));
}
if let Some(day) = input.strip_suffix('L') {
  return Ok(DaysOfWeek::Last(Field::DayOfWeek.parse_value(day)?));
}
```

So `7L` → `Last(6)` (last Saturday of the month), but standalone `L` → every Saturday. `0 0 * * L` therefore fires ~4–5× a month, not once. They are not equivalent — suggest dropping this sentence or documenting `L` as "every Saturday".

bartlomieju (commented):
Thanks for tightening this up — I checked the new content against Deno's actual parser (`ext/cron/cron.rs`) and the saffron crate it emulates, and the core corrections are right: numeric day-of-week really is `1=Sun..7=Sat`, `0` is rejected, and the `L`/`W`/`#` semantics plus the offset (1–30) and nth (1–5) bounds all match the source. Two of the added claims don't match runtime behavior, though — details inline.

runtime/fundamentals/cron.md
```diff
@@ -27,41 +27,101 @@ reference</a>
...
+- **N-th Day of Week:** `<day>#<nth>` triggers on the n-th specific day of the
+  week in a month. The `<nth>` value can be from `1` to `5`.
+
+### Examples
+
+| Expression              | Description                                               |
+| ----------------------- | --------------------------------------------------------- |
+| `0 9 * * 2-6`           | Every weekday at 09:00                                    |
+| `5/10 8-18 * * *`       | Every 10 minutes starting at :05, between 8 AM and 6 PM   |
+| `59-0 23-0 31-1 12-1 *` | From end of December to beginning of January (wraparound) |
```
bartlomieju: This range is wider than the description says. Deno's plain wraparound includes one value **before** the start (`range_bits`: `bit_range(start - 1, max) | bit_range(min, end)`, commented "Saffron includes one value before the start"). So `12-1` expands to months {11, 12, 1} — this also fires in **November**; likewise `31-1` → days {30, 31, 1}, `23-0` → hours {22, 23, 0}, `59-0` → minutes {58, 59, 0}. "From end of December to beginning of January" undersells it. Worth either picking a non-wraparound example here or spelling out the `start-1` quirk, since it's surprising.

runtime/fundamentals/cron.md
```diff
@@ -27,41 +27,101 @@ reference</a>
...
+| `0 0 L * *`             | Last day of every month at midnight                       |
+| `0 0 LW * *`            | Last weekday of every month at midnight                   |
+| `0 0 L-30W * *`         | Nearest weekday to (end of month - 30 days) at midnight   |
+| `0 0 1W * *`            | Nearest weekday to the 1st of the month at midnight       |
+| `0 0 * * 7L`            | Last Saturday of every month at midnight                  |
+| `0 0 * * MON#2`         | Second Monday of every month at midnight                  |
+| `*/10 20-4 * * *`       | Every 10 minutes during night hours (wraparound)          |
```
bartlomieju: Same `start-1` wraparound quirk as the December row above: `20-4` in the hour field actually expands to hours {19, 20, …, 23, 0, …, 4}, so this fires starting at 19:00, not 20:00.

runtime/fundamentals/cron.md
sanekb: My bad, won't make edits late at night anymore. Fixed it

runtime/fundamentals/cron.md
sanekb: Moved wraparound examples to the end and change description to show couple invocations in a `start-1` match

runtime/fundamentals/cron.md
sanekb: Updated the example to demonstrate that stepped wraparound ranges start at the literal start value

PR #3407, opened 2026-07-09
Visual regression fixes and improvements

josh-collinsworth:
A collection of visual, layout, and typographical fixes and improvements to address some regressions and side effects of recent updates. Make the colors a bit more standardized, and adjusts some of the color ramps for better consistency.

One outstanding item: it's highly inconsistent that *some* of our sidebar dropdowns behave like this (link and accordion button side-by-side):

CleanShot 2026-07-09 at 14 47 01

<p>&nbsp;</p>

...while others behave like _this_ instead (the whole thing is the accordion):

CleanShot 2026-07-09 at 14 54 00

<br />

I would prefer *one* approach to these dropdowns, not two (and I would suggest the latter, where a dropdown is never a page, is probably the best way to do it).


bartlomieju:
> I would prefer one approach to these dropdowns, not two (and I would suggest the latter, where a dropdown is never a page, is probably the best way to do it).

I agree on that for "CLI" and "Standard library", but items above (Running code, Dependency management, Testing, etc) I do think it makes sense that there's a top level page with high level overview/intro that has subpages. It's a lot better for SEO and linking from Google instead of having gigantic page with tons of anchors.

josh-collinsworth:
> > I would prefer one approach to these dropdowns, not two (and I would suggest the latter, where a dropdown is never a page, is probably the best way to do it).
> 
> I agree on that for "CLI" and "Standard library", but items above (Running code, Dependency management, Testing, etc) I do think it makes sense that there's a top level page with high level overview/intro that has subpages. It's a lot better for SEO and linking from Google instead of having gigantic page with tons of anchors.

I'm not necessarily proposing doing away with the pages themselves, for what it's worth. I just want the behavior to be consistent.

If we *do* keep the separate LP/accordion links as the convention, then we'd need to add landing pages for CLI and STD.

bartlomieju:
> If we do keep the separate LP/accordion links as the convention, then we'd need to add landing pages for CLI and STD.

I guess we can do that, it's that there would just not be a lot of content for these two. Let me come up with something.

bartlomieju:
@josh-collinsworth I opened: https://github.com/denoland/docs/pull/3408 would it help to unify everthing?

josh-collinsworth:
> @josh-collinsworth I opened: #3408 would it help to unify everthing?

@bartlomieju I'd say let's keep them separate. (Or just merge them both in sequence; either way.)

bartlomieju (approved):

PR #3408, opened 2026-07-09
docs: add CLI and Standard library sidebar landing pages

bartlomieju:
The "CLI" and "Standard library" groups in the runtime reference sidebar were
dropdown accordions with no clickable top-level page, so there was no landing
page introducing either section. The CLI subcommand overview already existed at
/runtime/reference/cli/ but the sidebar never linked to it.

This gives both groups the opt-in `disclosure` treatment used elsewhere in the
sidebar: the label becomes a link to the section landing page while a separate
chevron toggles the children. CLI now points at its existing overview. For the
Standard library, whose pages under runtime/reference/std/ are generated and
gitignored, the curated intro (what @std is, getting started with `deno add`,
and a note on package stability) is baked into the index template in
generate_std_docs.ts so it is produced on every build, with the package list
still auto-filled between the existing markers.


josh-collinsworth (approved):
Looks good, also added some styling to make the sidebar buttons look more like buttons

PR #3429, opened 2026-07-29
docs: add agents.md, an entrypoint for coding agents

bartlomieju:
Coding agents arrive at a Deno project carrying wrong priors: that Deno is a
separate ecosystem, that adopting it means rewriting imports or moving to JSR,
that `package.json` has to go. We already maintain agent skills that correct
all of this, but an agent only benefits from them once they are installed, and
nothing in the CLI or the docs currently tells it they exist.

So this page is deliberately an entrypoint rather than a guide. It orients the
agent, states the Node assumptions to unlearn, and then hands off to the skills
in denoland/skills, which stay the single source of truth for the command
surface, permissions, configuration layout, and migration. For agents that
cannot install skills it links the raw SKILL.md files, which are always
current, so nothing here needs restating and there is no second copy to drift.

The file lives in `static/`, so it is copied verbatim and served as
`text/markdown` by the existing markdown-source middleware. The URL we intend
to advertise is deno.com/agents.md; a companion dotcom PR serves this content
there, and a deno PR adds the link to `deno --help`.

### DIFF
--- a/static/agents.md
+++ b/static/agents.md
@@ -1,11 +1,5 @@
 # Deno for coding agents
 
-You are an AI coding agent. This page gets you oriented and then points you at
-Deno's agent skills, which hold the actual reference material. It is short on
-purpose — read it once, install the skill, and you will not need it again.
-
-## What Deno is
-
 Deno is a JavaScript and TypeScript runtime distributed as a single binary. That
 binary is also the package manager, formatter, linter, test runner, type
 checker, and compiler. It runs TypeScript directly, so a script needs no build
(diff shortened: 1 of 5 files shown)


static/agents.md
```diff
@@ -0,0 +1,128 @@
...
+This is the point of this page. Deno maintains agent skills that carry the full
+surface — dependency management, permissions, configuration layout, the built-in
+toolchain, publishing, and migration. They are what you should be working from,
+and they are kept current with the runtime.
+
+If the user agrees, install them. This touches no project source and adds no
+dependency:
+
+```sh
+npx skills add denoland/skills --skill deno
```
bartlomieju: If agent installs Deno as the first step this should probably say `dx skills ...`

bartlomieju: Good catch — fixed. Switched to `dx skills add denoland/skills --skill deno`, since section 1 has just installed Deno and `dx` ships with the binary, so this no longer assumes Node is present. Verified `dx skills` resolves and runs the CLI. Left `npx skills add ...` mentioned as the identical Node alternative.

